PR #3, opened 2020-05-15
explicit key,value based encoded selection feature ordering

akumar-cs:
- added key,value based segmented_feature selector for explicit ordering of idx->model
- tested by assigning wrong models to features and predictions sure dont match


notebooks/segmented_model_demo/cat_encoder.py
```diff
@@ -31,9 +31,9 @@ def __call__(self, x):
...
         return self._transformed_column_names
 
-    def cat_indexes_of_feature(self, feature: str) -> List[int]:
-        result = []
+    def cat_indexes_of_feature(self, feature: str) -> Dict[str,Dict[str,Any]]:
+        result = {}
         for idx, feat in enumerate(self._transformed_column_names):
             if feat.startswith(f"{feature}_"):
-                result.append(idx)
+                result[feat] = {'idx':idx, 'model': None}
```
sdraper-CS, line 38: I don;t think you should return a dict with a concept of `model` from this class because that is assuming a lter usage that may only be true for this context prefer to keep CatEncoder generic and we'll back-migrate the added capability to the one we use in other examples too).
Also I think the keys should the feature *value* not the encoded one-hot-col name (so the substring *following* `{feature}_`).  Hence in this case I'd expect a dict like:
```
{
    "SVM": 62,
    "logistic": 63
}
```

notebooks/segmented_model_demo/joint_model.py
```diff
@@ -1,18 +1,17 @@
...
+from typing import Dict
 
 import numpy as np
 
 
 class JointModel:
-    def __init__(self, segment_models: list, segment_selection_column_idxs: List[int]):
-        self.segment_models = segment_models
-        self.segment_selection_column_idxs = segment_selection_column_idxs
+    def __init__(self, segment_models_idx_mapping: Dict[str,Dict]):
```
sdraper-CS, line 7: Again, for genericity, I think `Dict[Any, Tuple[int, IHostedModel]]` and generate this dict from the one the encoder outputs in the notebook (since that's the glueing together of the generic components). For clarity expect an input dict keyed on feature value (`Any` or perhaps `Union[str,int]`) with values which are tuples of (encoded_column_index, segment_model) (`Tuple[int, IHostedModel]`)

sdraper-CS:
Addressed my comments above.  @akumar-cs - please could you review this branch now ;-)

akumar-cs:
> Addressed my comments above. @akumar-cs - please could you review this branch now ;-)

looks great. docstring( `Given value of the segmentation feature pick the segment model`) in the `get_seg_model` makes it much clear that feature value selection string is used to pick models and leaves no scope for user to fiddle around with model mapping.

akumar-cs:
@sdraper-CS this has both the commits for segmented models and xgboost model usecase, I added it all here(unfortunately though). includes fixes for StandardScaler in place of normalizer

notebooks/xgboost-model/xgboostDmatrixExample.ipynb
```diff
@@ -0,0 +1,558 @@
...
+    "        decoder(xg_reg.predict(xgb.DMatrix(encoder(X_test.values))))).all"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### creating a certifai evaluation dataset\n",
+    "\n",
+    "- earlier we modified our dataset to frop certain non useful columns\n",
```
cdraper-CS, line 381: frop => drop

notebooks/xgboost-model/xgboostDmatrixExample.ipynb
```diff
@@ -0,0 +1,558 @@
...
+    "- and we ran our encoder on the cleaned dataset\n",
+    "- we will pass the cleaned dataframe (with the removed columns) to certifai for evaluation\n",
+    "- this is needed since the dropped columns are non encoded and are completely useless for certifai evaluation"
```
cdraper-CS, line 384: It's needed because the model doesn't expect those columns, right?

notebooks/xgboost-model/xgboostDmatrixExample.ipynb
```diff
@@ -0,0 +1,558 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### wrapping model to create xgboost.Dmatix obj from numpy arrays for certifai predicts\n",
+    "\n",
+    "- cortex certifai invokes model (black-box) predicts using numpy-arrays from evaluation dataset provided\n",
+    "- since xgboost model requires Dmatrix obj for prediction we create a `TransformedPredict` wrapper class\n",
+    "- `TransformedPredict` wrapper creates Dmatix object before returning calling wrapped model's (here xgboost) predict"
```
cdraper-CS, line 208: Dmatix => Dmatrix here and in header

notebooks/xgboost-model/xgboostDmatrixExample.ipynb
akumar-cs: correct but even before that, these columns are not encoded. model also doesn't use the `outcome` or target column but we need that in the dataset for metrics calculation purposes

notebooks/xgboost-model/cat_encoder.py
```diff
@@ -0,0 +1,31 @@
...
+            self.normalizer = None
+
+    def __call__(self, x):
+        numeric = x[:, self.num_indexes]
+        if self.normalizer is not None:
+            numeric = self.normalizer.transform(numeric)
+        categorical = self.encoder.transform(x[:, self.cat_indexes]).toarray()
+        return np.concatenate((numeric, categorical), axis=1)
+
+    @property
```
sdraper-CS, line 29: Should probably use the same cat encoder (and in general try to standardize - we're very actively considering moving this into a toolkit utils library to avoid us having many slightly different versions proliferating all over the place).  In particular this one doesn't have the `cat_indexes_of_feature` method

akumar-cs: Makes sense,I’ll add the same encoder here as well

PR #5, opened 2020-05-21
sagemaker example using certifai model connector

akumar-cs:
- i have tested this only on sagemaker hosted notebook instance 
- for local running aws profile and role must be configured


akumar-cs:
@CognitiveScale/qa-bjesudasan-cs can you test the functionality and find any missing/unclear instructions

akumar-cs:
> Main notebook looks fine. I'm a bit confused by what is going on in `decode_data.py` though! Is there a SageMaker field naming requirement, or is this specifying. mapping to one-hot columns? Also I can't quite follow what's going on with `age` and `age_cat`

This is lifted from https://github.com/CognitiveScale/certifai/blob/master/certifai_engine/model_server/examples/german_credit/german_encoder.py. I added one extra line to stop outcome encoding

sdraper-CS:
> > Main notebook looks fine. I'm a bit confused by what is going on in `decode_data.py` though! Is there a SageMaker field naming requirement, or is this specifying. mapping to one-hot columns? Also I can't quite follow what's going on with `age` and `age_cat`
> 
> This is lifted from https://github.com/CognitiveScale/certifai/blob/master/certifai_engine/model_server/examples/german_credit/german_credit_schema.py I added one extra line to stop outcome column encoding

I'm still confused. In `decode_data.py` there's a big field mapping lookup dictionary beginning:
```
var_xwalk = {'checkingstatus':{
'A11' : '... < 0 DM',
'A12' : '0 <= ... < 200 DM',
'A13' : '... >= 200 DM / salary assignments for at least 1 year',
'A14' : 'no checking account'},
```
I don't see that in the example you linked above

bjesudasan-CS:
@akumar-cs I was able to run the notebook in Sagemaker successfully. Below are the observations.

1. Information on which conda environment to be selected in SageMaker is missing
(Selected `conda_python3`)

2. Had to install `sickit` and `numpy` explicitly

3. Had to add `import json` in `Test model predictions` step

4. Since I don't have access to the bucket that is provided in the notebook, had to change the S3 Bucket to the another to which I have access

Attached the notebook result for your reference

[CertifaiSageMakerExample.html.zip](https://github.com/CognitiveScale/cortex-certifai-examples/files/4667900/CertifaiSageMakerExample.html.zip)

notebooks/sagemaker_example/CertifaiSageMakerExample.ipynb
```diff
@@ -0,0 +1,1112 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Prerequisites\n",
+    "\n",
+    "This notebook was created and tested on an ml.m4.xlarge sagemaker notebook instance. However you can also run it locally by making sure you have the [aws profile](https://docs.aws.amazon.com/cli/latest/reference/configure/) configured and following dependencies installed on your system\n",
+    "\n",
+    "\n",
+    "- python>=3.6.2,<3.7\n",
+    "- scikit-learn=0.22.1\n",
```
cdraper-CS, line 25: = should be ==

cdraper-CS:
Is there any AWS doc link we can provide which helps user with understanding what they need for a "SageMaker-compatible role"?

cdraper-CS:
For running locally, `wget` is also a dependency (which I dont have on my Mac)

cdraper-CS:
Suggest you mention at the start of the notebook (as a prereq) that the user needs to have uploaded the certifai toolkit zip to an S3 bucket.

akumar-cs:
> Suggest you mention at the start of the notebook (as a prereq) that the user needs to have uploaded the certifai toolkit zip to an S3 bucket.

makes sense.

akumar-cs:
> Is there any AWS doc link we can provide which helps user with understanding what they need for a "SageMaker-compatible role"?

looking into this. should be there. although role i created during hosted instance creation included everything, i'll try to narrow this down

notebooks/sagemaker_example/CertifaiSageMakerExample.ipynb
akumar-cs: this was keeping conda in mind, since the default env on aws we use is conda. Maybe this is confusing since the test for numpy and scikit asks user to pip install. I'll fix these accordingly

akumar-cs:
> I'm OK with this, assuming you either get the local run to work or take that part out, and that it has been tested by one other person. Sorry, I havent been in a position to test it.

I was able to get it working locally by enabling TrustedPolicy for my dev user account in the execution iam-role i created for sagemaker. added docs for the same and ran certifai scan and fainess graph (output) is also attached to notebook. going ahead with the merge

PR #7, opened 2020-05-26
Fixed numeric encoding in CatEncoder

sdraper-CS:
See https://github.com/CognitiveScale/certifai/issues/2155


sdraper-CS:
Mostly because several model classes operate better with approx-Gaussian distributions rather than uniform ones.  MinMaxScaler in particular will not in any way normalize variance, so features with signal in small ranges are harder to pick up, whereas scaling them to equal variance tends to amplify those effects

PR #10, opened 2020-05-28
Segmented models

akumar-cs:
this is to merge segmented models and xgboost example to master
adding since min. of 2 reviews approvals required to merge this to  master


akumar-cs:
> Since this is all just moving stuff previously reviewed around, I'm ok with it on the basis you checked it all ran through once ok.

I haven’t run this branch yet, but I will before merging tomorrow morning

akumar-cs:
@sdraper-CS  with introduction `StandardScaler` there is a conversion warning that is emitted for all models now 
```
Data with input dtype object was converted to float64 by StandardScaler.
Data with input dtype int64 was converted to float64 by StandardScaler.
```

in `segmented models`, if i look closely at the jupyter-notebook `terminal` logs i see error logs pasted below even though notebook finishes without errors.

```

**Comment by sdraper-CS** [comment]
I think those warnings are fine, but if we really want to suppress them we should just add an `.astype(float)` at the numeric slice immediately before passing it to the scaler (on both `fit` and `transform`)

**Comment by akumar-cs** [comment]
> I think those warnings are fine, but if we really want to suppress them we should just add an `.astype(float)` at the numeric slice immediately before passing it to the scaler (on both `fit` and `transform`)

the error below with complete traceback shows up in certifai scan logs now, even though result seems unaffected. Should this be ignored ?
```
ValueError: Found array with 0 sample(s) (shape=(0, 64)) while a minimum of 1 is required. 
```

**Comment by sdraper-CS** [comment]
> > I think those warnings are fine, but if we really want to suppress them we should just add an `.astype(float)` at the numeric slice immediately before passing it to the scaler (on both `fit` and `transform`)
> 
> the error below with complete traceback shows up in certifai scan logs now, even though result seems unaffected. Should this be ignored ?
> 
> ```
> ValueError: Found array with 0 sample(s) (shape=(0, 64)) while a minimum of 1 is required. 
> ```

That looks more worrying.  Where is it coming from?  Do we have a dataset with 0 numeric components in it maybe?

**Comment by akumar-cs** [comment]
> > > I think those warnings are fine, but if we really want to suppress them we should just add an `.astype(float)` at the numeric slice immediately before passing it to the scaler (on both `fit` and `transform`)
> > 
> > 
> > the error below with complete traceback shows up in certifai scan logs now, even though result seems unaffected. Should this be ignored ?
> > ```
> > ValueError: Found array with 0 sample(s) (shape=(0, 64)) while a minimum of 1 is required. 
> > ```
> 
> That looks more worrying. Where is it coming from? Do we have a dataset with 0 numeric components in it maybe?

nopes, its the same segmented model usecase, dataset is just german_credit with appended cat column `model_selector` with SVM/Logistic as cat_values

**Comment by sdraper-CS** [comment]
Ah I see - we're not handling the case (in `JointModel`) where the particular batch has no examples in one segment or another.  I think the fix is is after the line:
```
seg_idxs = np.where(X[:, col_idx] > 0.5)[0]
```
add:
```
if len(seg_idxs) > 0:
```
to make the next two lines conditional

akumar-cs:
Done, tested merged

PR #12, opened 2020-05-28
Migrating toolkit examples

npasricha-cs:
All notebooks from toolkit brought in, tested with shared utils/datasets/definitions with new cat_encoder


npasricha-cs:
Do we want to keep all of the datasets/.json files? There are definitely a lot we can remove and I wanted to confirm which ones those should be. Can update so only the bare minimum needed to run the notebooks. Let me know your thoughts.

cdraper-CS:
@npasricha-cs The datasets, definitions and reports for the reference models are essential for the "getting started" experience and need to remain in the toolkit. We do not need to copy the reports into this repo. I would also only copy in the datasets that are essential to the notebooks.

npasricha-cs:
> @npasricha-cs The datasets, definitions and reports for the reference models are essential for the "getting started" experience and need to remain in the toolkit. We do not need to copy the reports into this repo. I would also only copy in the datasets that are essential to the notebooks.

Sounds good, I am not at the stage of removing anything from the toolkit just yet. I will keep only the essential files needed to run the notebooks successfully.

cdraper-CS:
I would not expect local_reports to be checked in. If it is in certifai repo, this is probably a mistaken checkin.

npasricha-cs:
> I would not expect local_reports to be checked in. If it is in certifai repo, this is probably a mistaken checkin. Should probably be added to .gitignore.

I believe local_reports are generated by the notebooks, so I will add that to the .gitignore

cdraper-CS:
Weren't we going to have a top-level datasets folder so we can share the key datasets needed across tutorials, notebooks and any other content that needs them?

npasricha-cs:
> Weren't we going to have a top-level datasets folder so we can share the key datasets needed across tutorials, notebooks and any other content that needs them?

I was under the impression we would have a top level datasets folder in notebooks and a top level datasets that remained in tutorials. I can adjust it to be shared across both tutorials and notebooks

npasricha-cs:
I have removed all extra datasets, definitions, added the local_reports folder to .gitignore, and changed the name of the sagemaker notebook for consistency with the others. All have been tested.

cdraper-CS:
Remove  `notebooks/local_reports` 

Typo in `notebooks/README.md` "explainaitons"

Would it make sense to leave the german credit definition that is specific to sagemaker in the sagemaker folder? I dont see it being likely to be used in other notebook examples.

npasricha-cs:
> Remove `notebooks/local_reports`
> 
> Typo in `notebooks/README.md` "explainaitons"
> 
> Would it make sense to leave the german credit definition that is specific to sagemaker in the sagemaker folder? I dont see it being likely to be used in other notebook examples.

WIll do, and that makes the most sense I think. @sdraper-CS, @akumar-cs what are your thoughts?

akumar-cs:
@npasricha-cs the [sagemaker-notebook](https://github.com/CognitiveScale/cortex-certifai-examples/blob/migrating_toolkit_examples/notebooks/sagemaker/SageMaker.ipynb) is not the one thats already there is master. Here is the [notebook](https://github.com/CognitiveScale/cortex-certifai-examples/blob/master/notebooks/sagemaker_example/CertifaiSageMakerExample.ipynb) in master,use this one.

PR #14, opened 2020-06-03
Added example for one-hot encoded datasets

sdraper-CS:
See https://github.com/CognitiveScale/certifai/issues/2039

Note that this is a PR to `pre-release` because this example requires the latest toolkit.


notebooks/one_hot_encoded/one_hot_encoded.ipynb
```diff
@@ -0,0 +1,939 @@
...
+    "        \n",
+    "# Save to an encoded CSV \n",
+    "df_encoded.to_csv(all_data_one_hot_file, index=False)\n",
+    "\n",
+    "# This is what the data looks like after encoding\n",
+    "display(df_encoded)\n",
+    "\n",
+    "# We will also want to know how the categorical encoding has mapped categorical values to columns, so\n",
+    "# save this as a dictionary for later use (if the data were given to us pre-encoded then we'd need\n",
+    "# to construct this mapping manually based on manual analysis of the data or a schem defintion also given\n",
```
npasricha-cs, line 671: schem -> schema

Appears an added "." as well

notebooks/one_hot_encoded/one_hot_encoded.ipynb
```diff
@@ -0,0 +1,939 @@
...
+    "#. to us with the dataset)\n",
```
npasricha-cs, line 672: Not sure if "." should be here

notebooks/one_hot_encoded/one_hot_encoded.ipynb
sdraper-CS: Fixed

notebooks/one_hot_encoded/one_hot_encoded.ipynb
sdraper-CS: Removed

cdraper-CS:
It would be useful to point out in the intro text that:
1. It's important to define the one-hot encodings and not just pass the data to Certifai, because otherwise Certifai will send the model invalid prediction requests
2. A benefit of this is that the results (e.g. explanations and I assume fairness burdens) are presented in terms of meaningful features/values. 

Did you test it works when fairness feature is one-hot encoded? Do  we need an example of that, either here or in doc?

Note this will need a PR on scan definition reference docs:
https://github.com/CognitiveScale/cortex-certifai/blob/master/docs/docs/reference/scan-definition.md

PR #19, opened 2020-07-23
Fixed missing property on SKLearn soft model wrapper

sdraper-CS:
SHould have been part fo the SHAP fix (meta) PR a couple days ago.  Change is pretty obvious!

### DIFF
--- a/notebooks/utils/sklearn_soft_wrapper.py
+++ b/notebooks/utils/sklearn_soft_wrapper.py
@@ -10,3 +10,6 @@ class SkLearnSoftWrapper(IHostedModel):
     def soft_predict(self, x):
         return self.skl_model.predict_proba(x)
 
+    @property
+    def supports_soft_scores(self):
+        return True


notebooks/utils/sklearn_soft_wrapper.py
```diff
@@ -10,3 +10,6 @@ def predict(self, x):
     def soft_predict(self, x):
         return self.skl_model.predict_proba(x)
 
+    @property
```
akumar-cs, line 13: this was working before without having this property. Why is it needed now?

akumar-cs: just looked at the issue 2464. and its fine. approved

sdraper-CS: The property was added with the comapper operator fusion handling in the optimization work so as to allow easy querying down the model wrapper stack

PR #20, opened 2020-08-04
german credit multiclass formulation and certifai example

akumar-cs:
notebooks for multiclass german credit 
- dataset creation
- one-hot and target encoded dataset persistence
- multiclass example notebooks including model train and certifai evaluation


sdraper-CS:
In the code to generate the 3 class model I noticed something that looks a little odd with the thresholding:
```
y_class_1 = y_preds[y_preds[:,0] > class_2_threshold][:,0]
y_class_1.shape
...
y_class_2 = y_preds[y_preds[:,1] > class_1_threshold][:,1]
y_class_2.shape
...
y_class_1 = y_preds[y_preds[:,0] > class_2_threshold][:,0]
np.sum(y_class_1 < class_1_threshold)
```
To determine class1 it uses the threshold `class_2_threshold` on output component [0]
To determine class2 it uses the threshold `class_1_threshold` on output component [1]
Then it overwrites `y_class1` (just confusing reuse of a variable name?) by taking the expression for the original `y_class1` (why recalculate this again BTW?) and comparing it to `class1_threshold`.  Thus `class1_threshold (which is used to determine y_class2!) is used on component t[1] of the output in one place and component [0] in another.

Aside from naming I think this could be made significantly simpler by adding a softmax over the two outputs after the last Dense (without the sigmoid activation since the softmax replaces it) so that you only have one soft output to deal with which is then thresholded with 2 cuts.  The current code that reuses one of the thresholds across the outputs doesn’t;t sea right because nothing enforces output1 + output2 = 1, which the softmax would (and then collapse it into a single value)

sdraper-CS:
Have you verified that the results that wind up being charted have reasonable correspondance to the results we normally get for German credit?  I think this would be a good sanity check to do (doesn't need to be shown in the notebook though)

sdraper-CS:
Given we have the target encoder instance would be good to also add some code to back-translate (in the case of an ambiguous result Ok to take the first for now - we'll do something better when we have code in the product for this) the target-encoded values and render the chart labels with the translations.  Net result should be charts that look extremely similar to the regular German Credit notebook results down to the labels also.  Since Prudential also have the pickled target encoder it's also pertinent to them as part of the example

cdraper-CS:
Please add a README that lets a user know what is being shown here (target encoding and multi-class). Will also need a link from docs.

cdraper-CS:
I tried to run generation notebook and got:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-1-2d2e3770897e> in <module>
     18 np.random.seed(seed)
     19 random.seed(seed)
---> 20 tf.random.set_random_seed(seed)
     21 
     22 base_path = '..'

AttributeError: module 'tensorflow._api.v2.random' has no attribute 'set_random_seed'
```

I have tensorflow 2.3.0, keras 2.4.3

cdraper-CS:
Running evaluation notebook I get:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-19-62f4e9cbb4ea> in <module>
     14 feature_upper_bounds = df_fairness[[f + ' upper bound' for f in feature_scores]]
     15 
---> 16 fig, ax = plt.subplots(figsize=[12,4])
     17 ax.set_title('status burdens by model', fontsize=20)
     18 colors = ['tab:blue', 'tab:orange', 'tab:green', 'tab:red', 'tab:pink']

AttributeError: module 'matplotlib' has no attribute 'subplots'
```

akumar-cs:
> df_fairness[[f + ' upper bound' for f in feature_scores]]

fixed. missing imports

akumar-cs:
> I tried to run generation notebook and got:
> 
> ```
> AttributeError                            Traceback (most recent call last)
> <ipython-input-1-2d2e3770897e> in <module>
>      18 np.random.seed(seed)
>      19 random.seed(seed)
> ---> 20 tf.random.set_random_seed(seed)
>      21 
>      22 base_path = '..'
> 
> AttributeError: module 'tensorflow._api.v2.random' has no attribute 'set_random_seed'
> ```
> 
> I have tensorflow 2.3.0, keras 2.4.3

i have Keras==2.3.1 and tensorflow 1.14.0, so the seed setting is different for diff version. i'll add try catch here

akumar-cs:
addressed review comments

sdraper-CS:
LGTM

cdraper-CS:
Typos in part2 notebook:

defintion
within the the environment

PR #22, opened 2020-08-11
updated description

akumar-cs:


cdraper-CS:
Approved, but still recommend we put in a branch not master

PR #23, opened 2020-08-14
Issue 2595

akumar-cs:
Added the `license.md` file along with the license header(with link to file) in each source file

ps: haven't added the script that adds these header licenses to the repo


akumar-cs:
> Looks good @akumar-cs. The blob links look a bit messy, but it also makes sense to go with something like that, so that we link to a specific version of that License file. It just means that if the file ever changes, we'll have to make sure to update the license file links as well.

correct. Also, like the absolute way to link a file is through the blob since without the blob(hash) git inserts `blob/<branch-name> ` in the link and then we have to preemptively insert `master` into the link and it would only show up when its merged to master .

akumar-cs:
merged

PR #24, opened 2020-08-20
Target encoding pro release

akumar-cs:
- all the work for target encoding example including deployment and remote scanning


cdraper-CS:
I think we should move this to be in its own examples folder, or if we leave it under "multiclass" then rename the second level to be "targed_encoded_example".   I still plan for us to add simple examples for the various types of multiclass (e.g. using iris, heart disease) at some point.

cdraper-CS:
Typos in part2 notebook:

defintion
within the the environment

cdraper-CS:
I'd suggest adding text similar to the following to the README to help a user know exactly what to do. Ideally use links to the notebooks not just the names. I've tested the first two parts but not the third (deploying/scanning in Azure)

Follow these steps to generate the target encoded multi-class dataset we will use in this example:

1. Follow instructions <here> to install the Certifai toolkit
2. Install additional prerequisites using `pip install keras tensorflow category_encoders matplotlib`:
3. Generate a multi-class dataset using the dataset_generation/german_credit_multiclass_dataset_generation.ipynb notebook
4. Generate a multi-class dataset using the dataset_generation/german_credit_multiclass_dataset_encoding.ipynb notebook

Follow these steps to run a Certifai scan for this model in a notebook:

1. Train the model using the certifai_multiclass_example/model_train_part1.ipynb notebook
2. Run a Certifai scan using the certifai_multiclass_example/certifai_multiclass_evaluation_part2.ipynb notebook

Follow these steps to move the model to Azure and scan it using Certifai Pro:

1. Deploy the model in Azure using the certifai_multiclass_example/deploying_model_part3.ipynb notebook
2. Install Certifai Pro in Azure and scan the model using the certifai_multiclass_example/remote_scan_part4.ipynb notebook

cdraper-CS:
Just noticed that in notebook 3 there are specific version requirements. It might be useful to help the user get those setup correctly up front.

Also this statement:
```
If you are using an Azure Machine Learning Notebook VM, you are all set. Otherwise, make sure you go through the configuration-notebook to create an Azure workspace. Creating remote environments/dependencies will be covered in the notebook
```

If I am using the VM, what step do I jump to? Assume I dont skip the whole notebook.

cdraper-CS:
The numpy version that the part 3 notebook says is needed (1.19.1) isnt supported by Certifai. What happens if you are using 1.18.x instead? That's what you'll get just installing Certifai in a clean conda env.

cdraper-CS:
I ran through part 3 then got stuck in part 4. I deployed a Certifai Pro. Then I got to:
```
Get the connection string for the storage account that holds the container called {az_container_name} in the cell below. You can obtain the connection string by following the Azure guide here.

Ensure that the storage account and container used here match the values you used at the beginning of this notebook to setup your Certifai Pro instance.
```

First, the Azure doc link has so much info that I dont think is relevant. I think the only relevant piece is that the connection string should look like:
`DefaultEndpointsProtocol=https;AccountName=storagesample;AccountKey=<account-key>`

Second, its not clear where to put this string. I assumed it was to replace `REDACTED`, i.e. as value of `az_credentials`. If it is, I think you should put `DefaultEndpointsProtocol=https;AccountName=<account-name>;AccountKey=<account-key>` instead of redacted and tell them to replace the account name and key.

I tried this with the storage I'd used for Certifai Pro, but I get `ResourceNotFoundError: The specified container does not exist.`. 
I think this is because I am trying to use a blob created for a different account or something.

cdraper-CS:
Got down to cell where set kubeconfig in certifai. The instructions need to say to edit the cell to replace  `certifai-kubeconfig.json` with the file you downloaded. Otherwise user just runs cell and wonders why it fails. Yup that's me.

cdraper-CS:
So the big gap at the end of this is - how does the data scientist now see the results?  

We at least need to point them at the Console.

_Possibly_ we need to help them retrieve the scan results and load them up into the API so they can run the visualizations we showed in the other notebook (like we did for HSBC). But let's discuss this next call to see if it's necessary.

cdraper-CS:
The last cell `!certifai remote describe 787b2ff6b9a4 --alias cpro-az` isnt useful because the user would have to figure out the scan id and to replace the `787b2ff6b9a4` in order to use it

cdraper-CS:
Another issue... I think the results are being written to `/<path>/az-pro-example/reports/german_credit_multiclass`

But the Pro I set up is looking for reports in  `/<path>/reports`. Anyway, user wont be able to see the reports in the Console.

akumar-cs:
addressed review comments

pkandarpa-cs (approved):

cdraper-CS:
Just a couple more suggestions from things I tripped over.

I missed this because its a comment in the code rather than an instruction before the cell. I dont tend to read thru all the code before I execute it.
`# replace the <certifai-kubeconfig.json> with the kubeconfig.json downloaded from the Certifai Console `

The other mistake I made was whenever I replaced the `<xxx>` I didnt put quotes in, just pasted in the replacement. So I think user will be less likely to mess up if you write it as `'<xxx>'`

cdraper-CS:
@akumar-cs If this is ready, can you merge it?

PR #25, opened 2020-08-24
Add step running preflight scan to CleanStart notebooks

laguirre-cs:


akumar-cs (approved):
Both notebooks look good. Slightly tangent question, does the `scan.run_preflight()` like a regular scan create  `preflight_scan_Id` analogous to `scan_id` everytime its run? if so, then what if there are multiple pre-flight scan ids, does the certifai scan then pick the latest one for estimates?

laguirre-cs:
> Both notebooks look good. Slightly tangent question, does the `scan.run_preflight()` like a regular scan create `preflight_scan_Id` analogous to `scan_id` everytime its run? if so, then what if there are multiple pre-flight scan ids, does the certifai scan then pick the latest one for estimates?

No, preflight scan's don't have anything analogous to a `scan_id`.  A single preflight report file will be written per model in the output directory, and rerunning the preflight scan will just update said files. When running the actual scan, we pick up existing preflight reports for each model (if they don't exist we create a new one).

PR #28, opened 2020-08-31
Add an example regression notebook

laguirre-cs:
See: https://github.com/CognitiveScale/certifai/issues/2697


cdraper-CS:
In the README, please mention that this example shows a preflight check. Probably as a keyword as well as in description

cdraper-CS (approved):
LGTM but please update readme to mention preflight

laguirre-cs:
> In the README, please mention that this example shows a preflight check. Probably as a keyword as well as in description

Addressed here: 35b7dcd

PR #29, opened 2020-09-09
Add example multiclass notebooks

laguirre-cs:
Issue: https://github.com/CognitiveScale/certifai/issues/2758


sdraper-CS (commented):
Comments on the notebooks:

Heart disease:
1. In the markdown at the top there is the phrase `and we will and run two scans` which seems mangled
2. In the scan creation all the outcomes have the same textual description (apart from stage) - is that correct?
3. We should consider adding a markdown cell immediately after the scan run to call out and comment on the warnings about sample sizes, so as to defuse potential questions
4. In the presentation of the explanations a little more white-space would help.  Suggest replacing he print statements for the titles (`Explanations for logistic Model:` etc) with markdown cells
5. Something that struck me in comparing the results was that the distribution of explainability for the logistic model was quite different between the two formulations.  Not sure if we should add a commentary markdown cell at the end to address this.  The reason (I think) is that when finding CFs for classes 2,3,4 it suffices (in the second formulation) to only change to the 'next' one in the ordering, whereas for the first formulation you have to move to the partition boundary (so from 4 all the way to 1 say) which will be harder.  The smaller moves may be more readily generated by fewer features changes, so that would explain the difference

Iris:
6. In the markdown preamble suggest 'Thusly' -> 'Consequently'

cdraper-CS (approved):
LGTM subject to Steve's comments

notebooks/multiclass/UnorderedMulticlass.ipynb
```diff
@@ -0,0 +1,505 @@
...
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "logistic Accuracy:  1.0\n",
+      "SVM Accuracy:  1.0\n"
+     ]
```
akumar-cs, line 467: both accuracy as 1 doesn't seem correct?

laguirre-cs:
> Comments on the notebooks:
> 
> Heart disease:
> 
> 1. In the markdown at the top there is the phrase `and we will and run two scans` which seems mangled
> 2. In the scan creation all the outcomes have the same textual description (apart from stage) - is that correct?
> 3. We should consider adding a markdown cell immediately after the scan run to call out and comment on the warnings about sample sizes, so as to defuse potential questions
> 4. In the presentation of the explanations a little more white-space would help.  Suggest replacing he print statements for the titles (`Explanations for logistic Model:` etc) with markdown cells
> 5. Something that struck me in comparing the results was that the distribution of explainability for the logistic model was quite different between the two formulations.  Not sure if we should add a commentary markdown cell at the end to address this.  The reason (I think) is that when finding CFs for classes 2,3,4 it suffices (in the second formulation) to only change to the 'next' one in the ordering, whereas for the first formulation you have to move to the partition boundary (so from 4 all the way to 1 say) which will be harder.  The smaller moves may be more readily generated by fewer features changes, so that would explain the difference
> 
> Iris:
> 6. In the markdown preamble suggest 'Thusly' -> 'Consequently'

1. Fixed

2. The reason for the duplication is that I haven't been able to find a better explanation for what the different values mean, beyond `0` meaning `< 50% diameter narrowing in any major vessel` and 1-4 meaning `> 50% diameter narrowing in any major vessel`.  The 1-4 labels are used in the dataset, but it is is usually analyzed as a binary-classification task for identifying the presence heart disease for this reason I think.

3. Added

4. Fixed

5. Added a small explanation mentioning the difference in explainability results for the logistic model specifically.

6. Fixed

sdraper-CS (approved):
LGTM

notebooks/multiclass/UnorderedMulticlass.ipynb
laguirre-cs: Good observation. I found this [SO post](https://stackoverflow.com/questions/36967126/why-do-i-get-good-accuracy-with-iris-dataset-with-a-single-hidden-node) where the first answer mentions a high accuracy is common because there is high correlation between two features and the outcome class (links to a file in the UCI archive verifying this).

Is this something that you think would raise some questions? (would you recommend commenting in notebook about this?)

akumar-cs: the code is absolutely fine, its just that the dataset is very simplistic with heavy correlation with the outcome columns. If you increase the test/train split ration to 0.4, accuracy comes down to 98.33, but i did it just to make sure. All looks cool to me.

akumar-cs (approved):
looks cool to me

PR #30, opened 2020-09-21
Add some examples of using the model SDK

cdraper-CS:


models/iris/train.py
```diff
@@ -0,0 +1,49 @@
...
+
+    from sklearn import datasets
+
+    # Bring in test and training data.
+    x, y = datasets.load_iris(return_X_y=True)
+    x_train,x_test,y_train,y_test=train_test_split(x,y,test_size=.5)
+
+    # Create an encoder
+    cat_columns = [] # All columns are numeric
+    encoder = CatEncoder(cat_columns, pd.DataFrame(x), normalize=True)
```
laguirre-cs, line 25: I think only the training dataset (`x_train`) is supposed to be passed to the encoder here, to avoid leaking any info about the test dataset (and invalidate the model accuracy below).

laguirre-cs (commented):
Mostly looks good, there just a good amount of minor formatting issues that should be addressed

models/iris/train.py
cdraper-CS: I double-checked with my resident expert, and its correct to pass the while dataset so that the encoder knows about all of the categories etc.

models/german_credit/README.md
```diff
@@ -0,0 +1,108 @@
...
+ The example shows how to use the Certifai Model SDK to wrap trained models into services
+ so they can be scanned using the
+ [Certifai toolkit](https://cognitivescale.github.io/cortex-certifai/docs/about).
+
+Specifically, it shows how to use the Certifai toolkit to:
+ * wrap a single model as a service
+ * implement a composite service that wraps multiple models
+ * scan the models
+
+## Wrap a single model as a service
```
smichalski-cs, line 16: The steps in the section below should be numbered.

models/german_credit/README.md
```diff
@@ -0,0 +1,108 @@
...
+Response from mlp/predict: [200] {"payload":{"predictions":[1]}}
+
+Response from svm/predict: [200] {"payload":{"predictions":[1]}}
+
+Response from health: [200] OK
+Response from shutdown: [200] Shutting down
+```
+At the end of the tests, the service is shutdown using the `shutdown` endpoint.
+
+## Scan models
```
smichalski-cs, line 88: These steps should also be numbered

models/iris/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+ The example shows how to use the Certifai Model SDK to wrap trained models into services
+ so they can be scanned using the
+  {Certifai toolkit](https://cognitivescale.github.io/cortex-certifai/docs/about).
+
+ Specifically, it shows how to use the Certifai toolkit to:
+  * wrap a single model as a service, using a customized wrapper
+  * scan the model
+
+## Wrap a single model using a customized wrapper
```
smichalski-cs, line 15: number the steps, please

models/iris/README.md
```diff
@@ -0,0 +1,66 @@
...
+```
+The tests create output similar to:
+```
+Response from model: [200] {"payload":{"predictions":[0,1,2]}}
+
+Response from shutdown: [200] Shutting down
+```
+At the end of the tests, the service is shutdown using the `shutdown` endpoint.
+
+## Scan model
```
smichalski-cs, line 52: more numbered steps here

models/iris/README.md
```diff
@@ -0,0 +1,66 @@
+# Iris classification example
+
+This example uses the [Iris plants dataset](https://scikit-learn.org/stable/datasets/index.html#iris-dataset)
+ provided with sklearn. This is a multi-class classification use case with no
+ favorability (outcomes are not favorable or unfavorable).
+
+ The example shows how to use the Certifai Model SDK to wrap trained models into services
+ so they can be scanned using the
+  {Certifai toolkit](https://cognitivescale.github.io/cortex-certifai/docs/about).
```
laguirre-cs, line 9: Last misformatted link! Just need to replace { with [

laguirre-cs (approved):
One last link to fix, but aside from that LGTM.

models/iris/README.md
cdraper-CS: aargh!

models/iris/iris_scanner_definition.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+# The `dataset_schema` section contains schema details about the datasets to be evaluated.
+dataset_schema:
+  outcome_column: 'species'
+  feature_schemas:
+  - feature_name: sepal_length
+    data_type: numerical-float
+    min: 4.3
+    max: 7.9
+    spread: 0.7000000000000002
```
akumar-cs, line 45: for numerical features, this is not necessarily needed as it is computed but doesn't do any harm to include

akumar-cs (approved):
looks good to go

smichalski-cs (approved):

PR #31, opened 2020-09-24
notebook that explores certifai fairness and robustness scores

jhenderson-cs:
This notebook trains models on simulated datasets and then explores how the fairness and robustness scores can change based on the model and the relationship between the target variable and the sensitive feature


cdraper-CS:
Please can you add a Copyright statement at the top - you should be able to copy it from another notebook

cdraper-CS (approved):
lgtm

sdraper-CS (commented):
Comments (can't do them in-line since I cannot view the rendering and insert comments there) [edit - Christine just pointed out to me that this is not stand-alone - its backing for the doc., so interpret/ignore some of my comments below as you see fit based on the assumption I made them with the mindset that this needed to stand alone too]:

1. `plot_decision_boundary ` and `plot_decision_boundary_cf` are almost exactly the same function - refactor most of this into one routine please

2. In the section where you display a representative point and its counterfactual for each value of alpha, I think in the spiel you should qualify 'distance' as L1-distance, otherwise the plotted points might strike someone as odd and cause them to think Certifai is getting bad approximations

3. I think the intuition on fairness needs a little more fleshing out.  The leap from 'disparate groups' have different distances to their counterfactuals to unfairness is a bit too large to leave unspoken.

4. Regarding robustness I think it comes across a bit too simply that robustness is just measuring overfitting, whereas the truth is more that they are correlated.  If all the robustness score did was tell you you have overfitting then the 'so what? - its easier for me to just hold out a validation set' is an obvious response.  Should probably (a) highlight that ONE of the things low robustness might be caused by is overfitting, and (b) point out that in ongoing monitoring this is useful because it does not require ground truth

5. The first mention of 'burden' is in the phrase 'We'll then examine the burden charts and the Fairness scores' - I think you need to define burden first!

jhenderson-cs:
I refactored the two plotting functions to be one and then added links to
the explanatory document to help make things clearer.

sdraper-CS (approved):
LGTM

PR #37, opened 2020-10-13
add soft scoring example,scan def and update readme

akumar-cs:
[#issue-2829](https://github.com/CognitiveScale/certifai/issues/2829)


cdraper-CS (approved):
LGTM but please also update model/README.md to highlight this example shows wrapping a model with soft scoring.

cdraper-CS:
@akumar-cs  I noticed when running the gunicorned model server that it creates a pkl file in current folder.  Is this going to cause issues if we need to get a gunicorned server running in UBI8/RHOS, given processes can only write to /tmp?

laguirre-cs (approved):
LGTM

akumar-cs:
> @akumar-cs I noticed when running the gunicorned model server that it creates a pkl file in current folder. Is this going to cause issues if we need to get a gunicorned server running in UBI8/RHOS, given processes can only write to /tmp?

yes, it will. We can only write to /tmp in rhos images. Since model sdk wasn't being run in container it wasn't an issue but ill see if i can save the state to /tmp might/not be possible due to references of class paths

akumar-cs:
> LGTM but please also update model/README.md to highlight this example shows wrapping a model with soft scoring.

Added `soft-scoring` part to  german_credit section of `models/README.md`.  Details for usage are inside the german_credit/README.md

PR #41, opened 2020-10-16
add artifacts for scoring dai pipelines

pkandarpa-cs:
Addresses Cognitivescale/certifai#2859

### DIFF
--- a/models/h2o_dai_german_credit/README.md
+++ b/models/h2o_dai_german_credit/README.md
@@ -92,7 +92,7 @@ Response from model: [200] {"payload":{"predictions":[1]}}
 
 ## Scan models using Certifai CLI
 
-1. Follow the instructions in the Certifai documentation under
+1. Follow the instructions in the [Certifai documentation](https://cognitivescale.github.io/cortex-certifai/docs/about) under
 'Toolkit > CLI Usage > Define and run scans locally' to define and test a
 scan definition for the German Credit example. The only difference from
 the example in the documentation is that you need to set the `predict_endpoint`


models/h2o_dai_german_credit/README.md
```diff
@@ -23,7 +23,7 @@ Specifically, it shows how to use the Certifai toolkit to do the following for t
 ## Usage Instructions
 
 ### Pre-requisites
-- Scoring Pipeline ([Mojo](http://docs.h2o.ai/driverless-ai/latest-stable/docs/userguide/scoring-pipeline-cpp.html#examples) or [Python](https://docs.h2o.ai/driverless-ai/latest-stable/docs/userguide/scoring-standalone-python.html#quick-start)) downloaded to disk on a Ubuntu 16.04 machine with Python 3.6. Refer to 
+- Scoring Pipeline ([Mojo](http://docs.h2o.ai/driverless-ai/latest-stable/docs/userguide/scoring-pipeline-cpp.html#examples) or [Python](https://docs.h2o.ai/driverless-ai/latest-stable/docs/userguide/scoring-standalone-python.html#quick-start)) downloaded to disk on a Ubuntu 16.04 machine with Python 3.6. Refer to the docs linked above to create and setup the environment to run this scoring pipeline
 
 ### Environment Setup
```
cdraper-CS, line 29: In the Environment Setup section and through the instructions, I suggest:
a) Tell reader what specific environment names we are going to use in the instructions, e.g. `certifai` for the scanner and `daimodel` for the prediction service.  Also tell them what each environment is for, i.e.  one for running scanner, one for running the prediction service.
b) Include the appropriate conda activate commands in the instructions

cdraper-CS:
How does a user recreate the german credit example in H2O Driverless AI? Don't we need to tell them what to do?

Re the input data being different for the two models, do we understand why and do we have confidence that someone trying to reproduce the example will be able to do so?

cdraper-CS (commented):
german_credit_mojo.csv is empty

cdraper-CS:
I thought I followed the python MOJO instructions, but when I try to run it I get:
```
(daimodel) ~/projects/cortex-certifai-examples/models/h2o_dai_german_credit: python app_h2o_mojo_pipeline.py
  File "app_h2o_mojo_pipeline.py", line 8, in <module>
    model = daimojo.model("./pipeline.mojo")
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/model.py", line 15, in __init__
    raise FileNotFoundError(filename)
FileNotFoundError: ./pipeline.mojo
```

models/h2o_dai_german_credit/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Environment Setup
+
+- Follow the instructions provided in the official H2O DAI documentation (linked above) to setup working python environments with all dependencies installed
+- Install the cortex-certifai-model-sdk and cortex-certifai-common packages into the same virtual environment / conda environment created by DAI's installation scripts for the scoring pipelines. 
+- We recommend that you install the full set of Cortex Certifai toolkit packages into a non DAI conda/virtualenv environment to prevent dependency conflicts.
+
+### Run the Certifai Wrapped DAI Model
+Please note that the downloaded scoring pipelines for MOJO and Python differ in the input data structure they expect for each predict call with the MOJO pipeline requiring `telephone` and `job` fields be present in the input data while the Python pipeline doesn't require the same. These changes are reflected in the two variants of the `germen_credit_{variant}.csv` files available in this folder where `variant` is one of [`mojo`, `python`].
```
cdraper-CS, line 35: germen => german
mojo csv is empty

cdraper-CS:
If the two models really need different datasets, then there needs to be two scan definitions. Currently it just refers to the one mojo dataset.

I also can't get the mojo to work. I replaced the empty dataset with the one from examples (including telephone and job) but I stil get an error from predict service:
```
2020-10-19 16:31:01,581 certifai.model.sdk.simple_wrapper ERROR    Model threw exception: "The names argument contains 18 elements, which is less than the number of columns being created (20)"
```

models/h2o_dai_german_credit/app_h2o_mojo_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+            'residence',
+            'property',
+            'age',
+            'otherplans',
+            'housing',
+            'cards',
+            'liable',
+            'job',
+            'telephone'
+        ]
```
cdraper-CS, line 36: these do not match the german credit dataset

models/h2o_dai_german_credit/app_h2o_python_scoring.py
```diff
@@ -0,0 +1,139 @@
+from certifai.model.sdk import SimpleModelWrapper
+import pandas as pd
+import numpy as np
+from numpy import nan
+import datatable as dt
+from scipy.special._ufuncs import expit
+from scoring_h2oai_experiment_7fa37640_0c78_11eb_9e3d_0242ac110003 import Scorer
```
cdraper-CS, line 7: The above import is specific to the downloaded pipeline. We need to figure out how to handle this from a "drop in" perspective and for the user who is trying to follow this example.

cdraper-CS:
Does the pipeline/MOJO have an API we can use to get the column names, rather than requiring them to be hardcoded?

cdraper-CS:
@pkandarpa-cs Suggest you focus on getting a clean MOJO example that a reader could recreate and execute, as that seems the most well behaved. Work on the python wheel once that is in place.

pkandarpa-cs:
> german_credit_mojo.csv is empty

I've added the content in

pkandarpa-cs:
> I thought I followed the python MOJO instructions, but when I tried to run it I got:
> 
> ```
> (daimodel) ~/projects/cortex-certifai-examples/models/h2o_dai_german_credit: python app_h2o_mojo_pipeline.py
> Traceback (most recent call last):
>   File "app_h2o_mojo_pipeline.py", line 8, in <module>
>     model = daimojo.model("./pipeline.mojo")
>   File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/model.py", line 15, in __init__
>     raise FileNotFoundError(filename)
> FileNotFoundError: ./pipeline.mojo
> ```
> 
> I was confused by there being both a runtime to download and a MOJO for the specific model. Maybe our instructions can clarify and also make it clear what have to do with the actual MOJO/python wheel.
> 
> Reading through again, I see that you say run it within the downloaded folder... but that will then require me to copy the example source there. Perhaps easier to say to copy the pipeline.mojo into the example folder. The python equivalent seems to be pip install the scoring experiment wheel (and the dependencies, but hopefully that's once off). I havent been able to test this because it only works in linux.

This should also work on OSX with the macOS whl for DAI runtime.

models/h2o_dai_german_credit/README.md
pkandarpa-cs: fixed empty mojo csv

models/h2o_dai_german_credit/app_h2o_mojo_pipeline.py
pkandarpa-cs: I'll change the README to differentiate between the two datasets. H2O DAI models don't work with the original dataset without dropped columns

models/h2o_dai_german_credit/app_h2o_python_scoring.py
pkandarpa-cs: Can use https://docs.python.org/3/library/importlib.html and ask the user to provide the name of the scoring `whl` directly

cdraper-CS:
Superceded by another PR - closing

PR #42, opened 2020-10-20
2850 dropin model

mchutani-CS:
This is a ~~draft~~ PR on the first pass on "drop-in" model utility.

Still needs:
- ~~Code generator to be written~~ Code generator has been written using Jinja2


models/containerized_model/README.md
```diff
@@ -0,0 +1,29 @@
+# How to create containerized models?
+
+## Pre-requisites
+
+- Certifai toolkit (from [CognitiveScale website](https://www.cognitivescale.com/try-certifai/)).
+- A model (pickle) stored in a cloud storage (e.g Amazon S3).
+- A base docker image which (strictly) simulates the trained model python environment.
```
cdraper-CS, line 7: I know what you mean by `(strictly) simulates` but we probably need to be more specific and say it has all of the dependencies at the specific versions that were used when the model was trained (or some such)

models/containerized_model/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Step 1 - Template generation
+TODO:
+- This should ask for full docker base image path (and inject into the Dockerfile) and generate the code template
+
+
+## Step 2 - Copy artifacts
+Copy the `packages` folder from inside the toolkit to current directory (`containerized_model` in this case)
+
+## Step 3 - Configure cloud storage
+- Add respective cloud storage credentials and `MODEL_PATH` to `environment.yaml` file.
```
cdraper-CS, line 19: Why/how would they be in conda environment.yml?  Isn't this more part of the run step

models/containerized_model/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Step 4 - Build and run
```
cdraper-CS, line 21: Separate build and run. Run is where you'd have to inject the right environment variables.

models/containerized_model/README.md
mchutani-CS: Okay I think I should rename the file to credentials.yml.
This is not a Fonda environment.yml but it’s a docker environment file which feeds the env variables at runtime.

models/containerized_model/README.md
mchutani-CS: Environment variables are being injected at runtime, but I just thought this is a cleaner way to put them in a file and then do a run instead of crunching so many env vars in the run command. Just syntactic sugar and looks neat. I’ll update the read me with more details and let me know if you think that would make more sense.

models/containerized_model/README.md
mchutani-CS: Update: I've still kept is as `environment.yml` since it does store all the environment variables to be passed to docker.
I've updated the readme so that it's clear that the `environment.yml` is being used at runtime.

cdraper-CS:
So I dont forget... we need a copyright on any source files (see https://github.com/CognitiveScale/cortex-certifai-examples/pull/45)

cdraper-CS (approved):

cdraper-CS (changes requested):
When I ran, I got:
```
ModuleNotFoundError: No module named 'certifai.model.utils.http_util'
```
And realized this is currently dependent on 1.3.6. Could you possibly put the code in line for now, otherwise Kaiser won't be able to get started for a couple of weeks.

models/containerized_model/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+```
+./container_util.sh build
+```
+
+## Step 5 - Run
+From inside the `<directory-name>` folder, run the following command which would run the docker image using environment variables from the environments file (`environment.yml`) that is being passed:
+
+```
+./container_util.sh run ./environment.yml
```
cdraper-CS, line 39: Default name of environments file so an just do `./container_util.sh run`

cdraper-CS:
I was imagining being able to test this with a local file (and I guess a file mount) before doing it with s3.  Worth adding instructions on how to do this?

models/containerized_model/README.md
```diff
@@ -0,0 +1,45 @@
...
+Add respective cloud storage credentials and `MODEL_PATH` to `<directory-name>/environment.yaml` file. This will be used in the `RUN` step.
+
+## Step 4 - Build
+From inside the `<directory-name>` folder, run the following command which would build the docker image:
+
+```
+./container_util.sh build
+```
+
+## Step 5 - Run
```
cdraper-CS, line 35: Run requires the model to be in place - need to add this to the instructions.

models/containerized_model/README.md
```diff
@@ -0,0 +1,58 @@
...
+For more `generate` options:
+```
+./generate.sh --help
+```
+
+## Step 2 - Copy artifacts
+Copy the `packages` folder from inside the toolkit into the generated directory `generated-container-model`:
+
+```
+cp -r <certifai-toolkit-path>/packages generated-container-model/
```
cdraper-CS, line 31: trivial thing but... if you change this to `cp -r <certifai-toolkit-path>/packages generated-container-model/packages` it's less prone to issues. When I first did this step I used Mac filename completion and ran cp -r <certifai-toolkit-path>/packages/ generated-container-model` which copied the contents of packages but not packages!

cdraper-CS:
Definitely smoother for me to follow this time. 

I encountered two things. 

First, I was trying to use this to wrap german credit model. So I ended up copying in the 'app_dtree.py' to become 'prediction_service.py'. I also copied in the pkl and just ran it from local file. That worked OK, but it did make me wonder if we shouldn't just have a 'src' folder in the generated dir, which contains the prediction_service.py but could have arbitrary source code copied into it (and which Dockerfile would copy over wholesale). 

I then tried to get the H2O model to run. This is obviously more complicated because it needs the downloaded daimojo wheel and a license file. I stopped at the point where I discovered that the linux daimojo doesnt work in the miniconda base image, so guess we will need a different base image. The 'delta' here is not just having to copy in additional files but also pip install additional content (the wheel plus datatable, protobuf). For first pass, this is probably just something we will need a custom Dockerfile for (plus instructions about copying over the wheel and I guess setting the license as an env var rather than baking it into image.

cdraper-CS:
For MOJO: 
1. missing step where user has to customize the prediction_service with the right columns and prediction classes.
2. Default MODEL_PATH for mojo template should be MODEL_PATH=/model/pipeline.mojo.  not model.pkl

I think the missing step should be inserted after Step 1. The way I'm thinking about this is...

Imagine there's a source repo for the model that has the train scripts etc. The devops or data scientists will run Step 1 to generate the 'generated-container-model' folder. They would then do the step to update the prediction_service.py (and add any other source needed). That would all be committed into the source repo, and pulled as part of the pipeline. 

Step 2 - 5 then all need to happen in the build pipeline. At the end, the pipeline will push the image somewhere.

cdraper-CS (approved):
Approving as I was able to follow and get working with local or s3,  but note comment about MODEL_PATH and adding step re customizing the prediction_service.py

cdraper-CS:
Was trying to follow an approach where I have a model repo and I want to generate the containerize template into it. I got an error - guess generate has to be run from the same directory:
```
(daimodel) ~/projects/cortex-certifai-examples/models/patient-readmission: sh ~/temp/cortex-certifai-examples/models/containerized_model/generate.sh -i patient-readmission:latest
Generating template using following configuration:
DIR = generated-container-model
BASE DOCKER IMAGE = continuumio/miniconda3:4.7.10
TARGET_DOCKER_IMAGE = patient-readmission:latest
MODEL TYPE = python
python: can't open file 'template.py': [Errno 2] No such file or directory
```

mchutani-CS:
Latest review comments have been addressed:
1. Added "update column" step as `Step - 2`
2. generate script works from anywhere now
3. Added seperate environment.yml for mojo and contains mojo related config

cdraper-CS:
In the h2o template prediction service, this is specific to German Credit (outcome values of 1 and 2):
```
def _get_prediction_class(preds):
    if preds[0] > preds[1]:
        return 1
    return 2
```

We at least need to add comments and possibly cause this function to throw unless filled in. It would also probably be more normal to return 1 if above threshold and 0 if below.

cdraper-CS:
Please include an entry for this folder in the [models/README.md table of contents](https://github.com/CognitiveScale/cortex-certifai-examples/blob/master/models/README.md#models-table-of-contents)

cdraper-CS:
why the change in german_credit from model to MODEL?

models/containerized_model/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+```
+cp -r <certifai-toolkit-path>/packages generated-container-model/packages
+```
+
+### Step 4 - Copy daimojo dependencies
+Copy `linux` - `daimojo` dependencies (`.whl` file) to `ext_packages` folder:
+
+```
+cp <path-to-linux-daimojo-file>.whl ext_packages/
```
cdraper-CS, line 49: Should be generated-container-model/ext_packages

cdraper-CS:
For MOJO, I am getting this error again:
```
ERROR: daimojo-2.4.8-cp36-cp36m-linux_x86_64.whl is not a supported wheel on this platform.
```
Looks like the Dockerfile has switched back to continuumio/miniconda3:4.7.10 ?

cdraper-CS (commented):
For MOJO, I am getting this error again:
```
ERROR: daimojo-2.4.8-cp36-cp36m-linux_x86_64.whl is not a supported wheel on this platform.
```
Looks like the Dockerfile has switched back to continuumio/miniconda3:4.7.10 ?

models/containerized_model/templates/environment.yml
```diff
@@ -0,0 +1,17 @@
+# An example for a model in s3 bucket
+#MODEL_PATH=s3://bucket/model.pkl
+
+# Default location for local model
+MODEL_PATH=/model/model.pkl
+
+# S3 credentials
+BUCKET_ENDPOINT=<your-s3-bucket-endpoint>
```
cdraper-CS, line 8: For BUCKET_ENDPOINT, I'd suggest commenting it as it only needs filling in if its something other than standard aws. Could you also give an actual example e.g. for aws?  I'm not actually sure what to put here

cdraper-CS:
Trying to run the prediction service on Mac outside of docker I get:
```
 in /Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/cppmojo.cpython-36m-darwin.so
(daimodel) ~/projects/cortex-certifai-examples/models/h2o_dai_german_credit: python prediction_service.py
  File "prediction_service.py", line 8, in <module>
    import daimojo.model
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/__init__.py", line 4, in <module>
    from daimojo.model import model
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/model.py", line 5, in <module>
    import daimojo.cppmojo
ImportError: dlopen(/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/cppmojo.cpython-36m-darwin.so, 2): Symbol not found: _EVP_MD_CTX_free
  Referenced from: /Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/cppmojo.cpython-36m-darwin.so
  Expected in: flat namespace
 in /Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/daimojo/cppmojo.cpython-36m-darwin.so
```

I found that reversing the imports so certifai imports happen first and the daimojo import happens last fixes this.

cdraper-CS:
See https://github.com/CognitiveScale/cortex-certifai-examples/pull/52 for proposed changes

mchutani-CS:
CLosing since it's done as part of https://github.com/CognitiveScale/cortex-certifai-examples/pull/52

PR #43, opened 2020-10-20
Add example for H2O Driverless AI

cdraper-CS:
This is specifically MOJO pipeline using german credit.

I have dropped the python wheel instructions for now, as its going to need more work to cleanly get something that is reproducible, although I've left the wrapper in the repo. 

Please can one of you run through these instructions  and check they are followable. You will need to sign up for a free trial at: https://www.h2o.ai/try-driverless-ai/ but easiest way to actually train model is to take the 2 hour trial at: https://aquarium.h2o.ai/

I also dropped the code in the wrapper that was reading the columns from the model. Problem with this is that this is the columns used by the model, NOT the columns in the input instances (or dataset). I don't see a way around those being provided in the wrapper right now. Looks like the H2O pipeline does its own filtering before actually calling the model, but to do that it needs an accurate list of the columns that are in the input instances.

Goal for the python wheel would be to have it able to use the same scan definition/datasets, but maybe just run the explanations scan. That's still 200+ minutes(!)


models/h2o_dai_german_credit/app_h2o_mojo_pipeline.py
```diff
@@ -0,0 +1,48 @@
...
+    def predict(self, npinstances):
+        instances = [tuple(instance) for instance in npinstances]
+        input_dt = dt.Frame(instances, names=columns)
+        predictions = model.predict(input_dt).to_pandas().apply(_get_prediction_class, axis=1)
+        return predictions.values
+
+if __name__ == "__main__":
+    # Host is set to 0.0.0.0 to allow this to be run in a docker container
+    app = GermanCredit(host="0.0.0.0")
+    app.run(log_level="warn")
```
akumar-cs, line 48: this is just ignored since its not a gunicorn server. no harm to keep it though

akumar-cs (approved):
tested the pr end-end it works

models/h2o_dai_german_credit/app_h2o_mojo_pipeline.py
cdraper-CS: The 'log_level'? That makes sense. I was _trying_ to get the noise out of certifai.log, but failed. Created a separate ticket to deal with that as it really makes it hard to check for scan errors.

mchutani-CS (approved):
LGTM. I'll test this out later tonight.

PR #46, opened 2020-10-22
Use CatEncoder from api  

cdraper-CS:
Cleanup a subset of notebooks to use CatEncoder from certifai common.

Also change test_user_case to test_use_case and remove hardcoded paths from scan definition as this is not really best practice

For xgboost, I also removed the install of the toolkit from the notebook. It will make it hard to test the notebook and thus maintain.

There are some extra columns in some outputs because I ran with 1.3.6, but I hope that is overlookable/future proofing!

Changed the license statements to refer to master rather than a particular blob @akumar-cs let me know if that was deliberate, but didnt seem to be what we'd want.


notebooks/xgboost-model/xgboostDmatrixExample.ipynb
```diff
@@ -19,44 +19,31 @@
...
-    "import sys\n",
-    "sys.path.insert(0, os.path.abspath(os.path.join('.')))\n",
-    "from cat_encoder import CatEncoder"
+    "from sklearn.model_selection import train_test_split\n",
+    "# imports for building certifai scan\n",
+    "from certifai.scanner.builder import (CertifaiScanBuilder, CertifaiPredictorWrapper, CertifaiModel, CertifaiModelMetric,\n",
+    "                                      CertifaiDataset, CertifaiGroupingFeature, CertifaiDatasetSource,\n",
+    "                                      CertifaiPredictionTask, CertifaiTaskOutcomes, CertifaiOutcomeValue)\n",
+    "from certifai.scanner.report_utils import scores, construct_scores_dataframe\n",
+    "from certifai.common.utils.encoding import CatEncoder"
```
akumar-cs, line 46: even tho this is imported from common lib but in cell 5 it again imports the encoder from cat_encoder file. need to remove that
```
# create encoder for categorical columns
from cat_encoder import CatEncoder
cat_columns = [
   'workclass', 
   'education', 
   'marital-status', 
   'occupation', 
   'relationship',
   'race',
   'gender',
   'native-country'
           ]
encoder = CatEncoder(cat_columns, X_train)```

Comment (akumar-cs) [reply notebooks/xgboost-model/xgboostDmatrixExample.ipynb:46]: need to delete the file `cat_encoder.py` and remove the re-import in `cell5`

File: notebooks/xgboost-model/override_model_predict.py [hunk 2]
```diff
@@ -0,0 +1,9 @@
+
+import numpy as np
+import xgboost as xgb
+class TransformedPredict:
+    def __init__(self,model):
+        self.model = model
+    def predict(self,arr):
+        dtest = xgb.DMatrix(data=arr)
+        return self.model.predict(dtest)
```
akumar-cs, line 9: this file should not be checkin, its created in the notebook itself using cell magic, for the users to have see why its needed and not have to open another file. it needs to be separate from notebook coz of the multiprocessing inside notebook issue. We should either keep the file and remove the %%wite override_model_predict.py from the cell or just leave the magic and not check in the file

notebooks/xgboost-model/decoder.py
```diff
@@ -0,0 +1,10 @@
+import numpy as np
+
+class Decoder:
+    def __init__(self,threshold):
+        self.threshold = threshold
+    
+    def __call__(self,x):
+        if not isinstance(x, np.ndarray):
+             x = np.array(x)
```
akumar-cs, line 9: same as below comment on`override_model_predict.py` file

akumar-cs:
i kept it as a blob as it was in a branch before it was merged to master and in that case it show as broken link since it was being reviewed and not in master. but now its in master so it should be fine. infact even if the other links in other files are not updated to point to branch master by name it would still work since it refers to the hash which is permanent.

tldr:  license change to point to master is fine

akumar-cs (approved):
apart from the two comments on xgboost example rest lgtm

mchutani-CS (approved):

mchutani-CS (approved):

akumar-cs (approved):

laguirre-cs (approved):
LGTM

PR #48, opened 2020-10-23
Use construct_explanations_dataframe in multi class notebooks

cdraper-CS:
Use construct_explanations_dataframe in multiclass notebooks. Plus other small cleanup.


akumar-cs (approved):
Looks cool

sdraper-CS (changes requested):
The explanations are broken and/or very co fusing for the regression model

laguirre-cs:
Two things:

1. I think the explanations in the regression notebook are missing the changed columns

2. It might also be a nice touch in the ordered scan of the `MulticlassHeartDisease.ipynb` to have the explanations show an instance where 2 counterfactuals were generated. It looks like both of the chosen observations have an initial prediction of 0, so we can't find a "more favorable" counterfactual

cdraper-CS:
@laguirre-cs  Fixed the problem with regression notebook (was dropping columns with any 'N/a' not all 'N/a'). 

On the multi-class notebook, issue was actually two bugs. display_explanations was only showing the first change, and it was using the partitioned results not the ordered results! So thank you again for good catches -  I'm afraid I went 'notebook blind' on Friday. Also found a couple of places in original code where it was using wrong results so fixed those too.

sdraper-CS (commented):
MulticlassHeartDisease.ipynb:

1.  (not new but...) Opening markdown cell is ungrammatical `...framing the use case as a partitioned and ordered multiclass classification tasks respectively` - singular and plural

2. I think we don;t actually know any better due to poor source material but it continues to bug me that we have:
```
        CertifaiOutcomeValue(1, name='Stage 1: > 50% diameter narrowing in a major vessel'),
        CertifaiOutcomeValue(2, name='Stage 2: > 50% diameter narrowing in a major vessel'),
        CertifaiOutcomeValue(3, name='Stage 3: > 50% diameter narrowing in a major vessel'),
        CertifaiOutcomeValue(4, name='Stage 4: > 50% diameter narrowing in a major vessel'),
```
(all unfavorable outcomes have the same descriptive text).  Perhaps we should just drop the text after `Stage <n>` in all or all but the first case?

3. In the notes markdown cell where we talk about the warnings we say `not enough samples of each protected class within the sex feature for Certifai's genetic algorithm to converge towards the optimal counterfactuals`.  This is inaccurate - the small sample size does not (primarily) impact the GA .  It impacts the convergence of the aggregate statistics that derive from the individual GA results

4. Obviously its deliberate that you show explanations for different individuals for the two different models, but its very counter-intuitive (I would naturally want to look at what the models explained differently).  It either needs different handling (e.g. - do two rows for each model if its a diversity reason) or an explicit explanation in a markdown cell as to why you're doing it or else it will result in confusion

5. Why is 'row==1' a different person between the two formulations (ordered and partitioned)?  I really wanted to compare the changes that the choice of formulation resulted in

Regression.ipynb:

6. We're inconsistent about running the pre-flight.  We only do it in this notebook (just saying - no need to change it now!)

7. Um...is it really a good example to show a negative claim settlement amount?  Actually it *may well be* since this is a weakness of the *model* that is exposed by Certifai, but it really needs a markdown cell discussing it! (final explanation)

laguirre-cs (approved):

cdraper-CS:
@sdraper-CS Latest changes address review comments 1 to 4. Not planning to take action on 6 or 7 at this point. Will rerun to address 5 when the code changes are merged.

cdraper-CS:
Have rerun notebooks and pushed update to multi-class now the fixes are in

PR #52, opened 2020-10-30
Containerizing models

cdraper-CS:
Updated PR for my latest comments to https://github.com/CognitiveScale/cortex-certifai-examples/pull/42

I squashed some commits in order to remove the unintentional german credit model changes.

Main other changes:
* Changed H2O prediction service so that user has to fill out _get_prediction_class
* Changed H2O prediction service so that service can run outside of docker and picks up local files if MODEL_PATH not defined
* Changed order of imports to avoid issue when running outside of docker on Mac
* Reordered Dockerfile so user can change prediction service without having to rebuild all layers
* Updates to README for flow
* Commented out BUCKET_ENDPOINT


mchutani-CS (approved):
LGTM. Tested on both (mojo and python)

akumar-cs (approved):

PR #54, opened 2020-11-02
read_and_save_model typo

akumar-cs:
read_and_save_file doesn't exist in this module, perhaps in mojo service template.

i think last night changes introduced the typo

### DIFF
--- a/models/containerized_model/templates/prediction_service.py
+++ b/models/containerized_model/templates/prediction_service.py
@@ -35,7 +35,7 @@ def main():
     if model_path is not None:
         # Copy files from remote path to local
         default_model_path = os.path.normpath(os.path.join(CURRENT_PATH, '../model', os.path.basename(model_path)))
-        read_and_save_file(model_path, default_model_path)
+        read_and_save_model(model_path, default_model_path)
     else:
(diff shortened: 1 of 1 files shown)


cdraper-CS (approved):

models/containerized_model/templates/prediction_service.py
```diff
@@ -35,7 +35,7 @@ def main():
     if model_path is not None:
         # Copy files from remote path to local
         default_model_path = os.path.normpath(os.path.join(CURRENT_PATH, '../model', os.path.basename(model_path)))
-        read_and_save_file(model_path, default_model_path)
+        read_and_save_model(model_path, default_model_path)
```
mchutani-CS, line 38: We should probably change the function name to `read_and_save_file` for consitency since `prediction_service_h2o_mojo` has it named as `read_and_save_file`. Check here: https://github.com/CognitiveScale/cortex-certifai-examples/blob/master/models/containerized_model/templates/prediction_service_h2o_mojo.py#L36

mchutani-CS (changes requested):
Made a minor comment.

mchutani-CS (approved):

PR #57, opened 2020-11-02
2916 remove workaround from patient readmission

cdraper-CS:
I removed the encoder that is not needed after the 1.3.6 bug fix.
Also changed the logit model to 'liblinear' so it converges.
Added explanation of why there are warnings when scan def is saved/loaded.


notebooks/patient_readmission/patient-readmission-trust-results.ipynb
```diff
@@ -135,25 +140,25 @@
        "    <tr style=\"text-align: right;\">\n",
        "      <th></th>\n",
        "      <th>context</th>\n",
-       "      <th>type</th>\n",
-       "      <th>overall fairness</th>\n",
+       "      <th>type in []</th>\n",
```
laguirre-cs, line 143: The column's name changed from `type` (e.g. burden) to `type in []`. Something seems wrong here because that column name is off and there are now multiple columns named `type in Feature (age)` that all have the value of "burden"...

Also, the column count jumped from 24 to 42 with this re-run (because of all the new `type in Feature (age)` columns), so columns like `Group details (AfricanAmerican)` aren't displayed.

laguirre-cs (commented):
Two things:

1. the fairness scores table in the trust results notebook seems off (see my comment)

2. There's a broken link in the trust-results notebook pointing to the fairness metrics notebook example (it's under the table mentioned in (1))

notebooks/patient_readmission/patient-readmission-trust-results.ipynb
cdraper-CS: @laguirre-cs For (1), I'd forgotten about those undesirable side effects in construct_scores_dataframe. It now really doesnt work great for fairness. I couldnt find a clean way to fix it without lots of special casing. Anyway, I changed the example to use the fairness-specific functions. 

For (2), fixed

laguirre-cs (commented):
The changes to the readmission example look fine to me. But please remove all the JSON report & unrelated files before merging!

laguirre-cs (approved):

PR #59, opened 2020-11-05
Issue 2911

akumar-cs:
pr against master to compare changes with containerized templates ,once design is validated will. marking pr as draft for that purpose


models/h2o_dai_regression_auto_insurance/app_h2o_mojo_pipeline_gunicorn.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        overridden method to make external global imports
+        When using external imports override the method to provide necessary imports. Make sure to mark them `global` to
+        be used by certifai interpreter correctly. set once,use throughout.
+        :return: None
+        """
+        global dt
+        import datatable as dt
+
+    def _get_prediction_class(self, preds):
```
cdraper-CS, line 53: I think we should rename the function to be _get_prediction given what it does in the regression case is return a probability not the class

models/containerized_model/templates/prediction_service_h2o_mojo.py
```diff
@@ -39,27 +30,51 @@ def read_and_save_file(source_path, destination_path):
...
+        overridden method to make external global imports
+        When using external imports override the method to provide necessary imports. Make sure to mark them `global` to
+        be used by certifai interpreter correctly. set once,use throughout.
+        :return: None
+        """
+        global dt
+        import datatable as dt
+
+    # <Fill in this function and remove the 'raise'>
+    def _get_prediction_class(self, preds):
```
cdraper-CS, line 62: It's highly desirable to have a "will usually work for H2O" implementation here, and have the user override it only if needed using a list of the outcome classes in the order they will be returned (c.f. columns).   We can then move the overrides into metadata that's optionally provided with the model but that'll need more refactoring of containerization scripts so perhaps leave that out of this PR.

Also think we should rename the function to be _get_prediction given what it does in the regression case is return a probability not the class

cdraper-CS (commented):
Do any of these changes require 1.3.6, or is that only when the user turns on production=True?

models/containerized_model/templates/prediction_service_h2o_mojo.py
akumar-cs: all these are 1.3.6. like gunicorn won't work with 1.3.5  if any of the models require external imports like xgb and datatable (in-case of h20). I can add that docs specifying these examples work with 1.3.5 and we should keep this in a separate branch for now and merge as and when required

akumar-cs: _get_prediction for regression returns actual prediction values like a floating point  number not probabilities, so yea i should rename it to get_prediction

akumar-cs: h2o that works out of box will require us to either fix deafult case as binary classification or accept metadata file

cdraper-CS: @akumar-cs Why can't we figure it out from the preds (1, 2 or more results per...)?

cdraper-CS:
Have incorporated these changes into other PRs and merged work for xgboost and H2O regression. Remaining open one is: https://github.com/CognitiveScale/cortex-certifai-examples/pull/67

cdraper-CS:
Closng as superceded by other PR

PR #61, opened 2020-11-06
2910 deploy to k8s

mchutani-CS:
~~Draft PR for easy collaboration on improvements.~~
PR for pipeline 2


models/containerized_model/template.py
```diff
@@ -21,7 +21,46 @@ def create_directories(directory_list: list):
             except FileExistsError:
                 pass
 
-    def generate_python():
+    def generate_base_file_metadata():
+        image_name, _ = args.target_docker_image.split(':')
+        deployment_name = image_name.split('/')[-1].strip()
```
cdraper-CS, line 26: One think I hit. This name can only be alphanum or "-". I had an image name with "_" and the deployment fails. Maybe we should make deployment name a param with a default. Or make generating the deployment a separate generate action where you give it a name. The latter would mean we could deal with case where there's a generic image (e.g. for scikit models, or for h2o models) but multiple deployments.

mchutani-CS: Have made those changes now. I'm not sure where to consolidate this information in the README, will leave that decision to you.

models/containerized_model/DEPLOYMENT.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Update Scan definition file with valid endpoints
+Update `predict_endpoint` field under `models` section in your `scan_definition.yaml` file to the service endpoint. The service endpoint can be created as follows:
+
+`http://<service-name>.<namespace>.svc.cluster.local:8551/predict`
+
+`<service-name>` : should be the same as `metadata.name` under `kind:Service` in `generated-container-model/deployment.yml` file.
+`<namespace` : `containermodel` (default namespace)
+
+### SSH into the bastian host
```
cdraper-CS, line 73: spelling bastian => bastion (*2)

cdraper-CS (approved):

models/containerized_model/DEPLOYMENT.md
```diff
@@ -0,0 +1,90 @@
...
+          value: "<your-s3-secret-key>"
+        - name: BUCKET_ACCESS_KEY
+          value: "<your-s3-access-key>"
+```
+
+## Step 3 - Deploy
+Run the following command in order to deploy the resources mentioned in the `generated-container-model/deployment.yml` file:
+
+```
+kubectl apply -f generated-container-model/deployment.yml
```
bjesudasan-CS, line 53: namespace provided in the above manifest should exist else running this command would result in an error.

models/containerized_model/DEPLOYMENT.md
```diff
@@ -0,0 +1,90 @@
...
+spec:
+  containers:
+    - image: "<your-docker-image>"
+      imagePullPolicy: IfNotPresent
+      name: {{DEPLOYMENT_NAME}}
+      ports:
+        - containerPort: 8551
+          protocol: TCP
+      env:
+        - name: MODEL_PATH
```
bjesudasan-CS, line 39: deployment file has placeholdrs for only S3 and not for azure blob or gcs bucket, is it expected?

models/containerized_model/DEPLOYMENT.md
```diff
@@ -0,0 +1,90 @@
...
+kubectl apply -f generated-container-model/deployment.yml
+```
+
+This command will deploy a `Service` and a `Deployment` under the namespace specified in the `generated-container-model/deployment.yml` file (`containermodel` by default).
+
+
+## Step 4 - Running a remote scan
+
+### Pre-requisites
+- datasets have to be in cloud storage (`s3`) and accessible by the same credentials that the scanner uses for writing reports; scan definition needs to have this url
```
bjesudasan-CS, line 62: `scan definition needs to have this url` - Are we referring to the model endpoint? if yes, can we add some more context to it

models/containerized_model/DEPLOYMENT.md
```diff
@@ -0,0 +1,90 @@
...
+
+`http://<service-name>.<namespace>.svc.cluster.local:8551/predict`
+
+`<service-name>` : should be the same as `metadata.name` under `kind:Service` in `generated-container-model/deployment.yml` file.
+`<namespace` : `containermodel` (default namespace)
+
+### SSH into the bastion host
+This step is only needed if you don't have access to kubernetes from your local machine. SSH into the bastion host which has access to the k8s cluster.
+
+If you have local access, then install certifai packages from `packages/all` folder inside the toolkit.
```
bjesudasan-CS, line 76: Below two pip commands are misleading as if installing these two packages is sufficient.

i feel its better to provide the certifai toolkit setup docs link instead - https://cognitivescale.github.io/cortex-certifai/docs/toolkit/setup/install-certifai-cli-lib

models/containerized_model/DEPLOYMENT.md
mchutani-CS: I think we've kept it since the customer uses Minio. I'll let @cdraper-CS comment if we need to add other stuff as well.

bjesudasan-CS (approved):
LGTM

PR #63, opened 2020-11-09
2911 xgboost example

cdraper-CS:
Adds example using XGBoost that works with Certifai 1.3.5, based on example created by @akumar-cs 

Has a number of workarounds for issues with Shap evaluations:
* Eval + explanation dataset rows must be less than 10000, so create a smaller eval dataset after training
* Need to return predictions for both class labels and not use a single label + threshold
* Suppress high-cardinality categorical encoding pending Shap target encoding fix


bjesudasan-CS (approved):
Able to run through the tutorial.

models/income_prediction/train.py
```diff
@@ -0,0 +1,79 @@
...
+    threshold = 0.46
+    dtest = xgb.DMatrix(encoder(X_test.values))
+    preds = xg_reg.predict(dtest)
+    best_preds = map(lambda x: int(x > threshold), preds)
+    acc = accuracy_score(y_test, list(best_preds))
+    print(f'model accuracy on test set: {acc}')
+
+    # dump model and encoder to disk
+    filename = 'adult_income_xgb.pkl'
+    model_obj = {
```
sdraper-CS, line 64: I think it would be best if the threshold were to be added to the pickled dict, so that the 0.46 doesn't appear in two places

sdraper-CS (approved):
Ok to merge, though I have a minor comment (but not a blocker)

models/income_prediction/train.py
cdraper-CS: Oh. I'm not currently passing the threshold into the wrapper, which I think means its using 0.5. I need to fix that.

PR #69, opened 2020-11-19
Fast explain example

sdraper-CS:
See issue #2971 in the Certifai repo


cdraper-CS:
I updated pre-release to be equal with master in the hope that this would remove the 26 spurious commits, but it didnt. You may need to rebase.

cdraper-CS:
Please add to model README

cdraper-CS:
If this is intended to merge at some point, it'll need some words :-)

Need to tell user that it has category_encoders as a prereq (pip install)

cdraper-CS:
"one-tim" compute.  Where's bert and harry?

cdraper-CS:
The notebook runs fine for me

laguirre-cs (commented):
I'm only commenting on the fast explanations notebook:

* Think Christine already found all the misspellings
* The `_wrap_sklearn_classifier` seems a little redundant (also the `CertifaiPredictorWrapper` was already imported in cell 1)
* The notebook looks fine to me, and I could re-run it problem
* (This is really a side comment) I noticed this line in the notebook: `pd.set_option('display.max_columns', None)`. I think we should be setting this value in notebooks whenever we display explanations. I know there are a couple examples that have truncated tables (e.g. regression notebook)

sdraper-CS:
@cdraper-CS @laguirre-cs Addressed all review comments I think

laguirre-cs (approved):
(Again only looking at the notebook and related README entry)

Needs a small update for the README entry, but notebook looks good. I think a markdown cell comparing speed up would pack a good punch at the end (but I'm assuming it's still too early to try and do that comparison). So, i'm approving w/ the caveat of the README update.

cdraper-CS (commented):
I checked this runs without needing the beta flag, so I suggest you merge to master but please add a "Requires Certifai version 1.3.7+" prereq.

Found two more typos: exaplantions, neccessary

sdraper-CS:
@cdraper-CS Addressed your final outstanding comment

cdraper-CS (approved):

PR #74, opened 2020-11-25
V2 Template changes

cdraper-CS:
Apologies for the big PR.  Main thing from a review perspective is (1) overall approach (2) can you successfully run the test script (3) anything obvious I've messed up for the remote deployment - I had a fairly nasty rebase to do.

If you have problems with xgboost models, its probably a version mismatch. The docker container has xgboost==1.2.0.  

For V2 templates:

Restructured all the templates by model type, to make it easier to add xgboost and others.

Added requirements.txt and metadata.yml

Added examples of containerization plus a test script that generates, builds, runs and tests them all, including reusing the same image across multiple models. You do need a valid H2O license to run the scripts (or comment out the H2O models in run_test.sh).

Refactored prediction service to break out utils (just the things that are done outside of the wrapper, as unfortunately I dont think we can factor out a wrapper class unless we put it into certifai). To allow this, I had to change the docker entrypoint or it didn't pick up utils.py. We probably need to come back to how we can allow additional source files.

On xgboost, I realized that there is a difference between:
* XGBClassifier or XGBRegressor - behaves just like sklearn model
* xgboost using DMatrix - a lower-level more performant predict, but probably only if you want the probabilities

I'm proposing we should focus on the former, but I didnt want to lose the work on the latter. So to this effect, I added an xgboost iris example using XGBClassifier which uses the base python template (but has to update requirements and set some flags in metadata). And then a separate python-xgboost-dmatrix wrapper, which currently should work for binary classification and regression but I don't intend to spend effort on getting that to work for multi-class in the near term.

Work on making the deployment template more generic will be a separate PR.


models/containerized_model/DEPLOYMENT.md
```diff
@@ -38,6 +39,8 @@ spec:
       env:
         - name: MODEL_PATH
           value: "s3://bucket/model.pkl"
+        - name: METADATA_PATH
+          value: "s3://bucket/metadata.yml"
         - name: BUCKET_ENDPOINT
           value: "<your-s3-bucket-endpoint>"
```
akumar-cs, line 45: is the bucket endpoint generic for diff. cloud providers like azure or limited to S3 at this time?

cdraper-CS:
Going to merge into v2-templates branch and just treat that as a working feature branch for now

PR #75, opened 2020-12-01
Add xgboost iris example

cdraper-CS:
Add an xgboost example that follows sklearn pattern rather than using raw dmatrix


sdraper-CS (commented):
One minor comment, but basically fine.  Feel free to merge as is if you prefer

laguirre-cs (approved):
I think you can also update the models README to list xgboost under "Model Frameworks" for the iris example. Aside from that, LGTM

PR #77, opened 2020-12-04
Specify numeric features for patient readmission

cdraper-CS:
Updated the explain-scan notebook to specify integer features that otherwise may be inferred as categorical; then reran all notebooks.


sdraper-CS (approved):
Code change LGTM, so approved so long as you are not seeing anything that bothers you in the results

PR #82, opened 2020-12-23
clean up xgboost dmatrix v2-template

akumar-cs:
issue-https://github.com/CognitiveScale/certifai/issues/2950

### DIFF
--- a/models/containerized_model/README.md
+++ b/models/containerized_model/README.md
@@ -158,7 +158,7 @@ in your current directory with the generated code.
 
 This template is designed to work with a standard scikit-learn model,
 and with an XGBClassifier or XGBRegressor model. For an xgboost model using
-DMatrix, use `-m python-xgboost-dmatrix`.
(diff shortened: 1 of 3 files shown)


cdraper-CS:
I think what you've done makes a lot of sense (force soft for classification, hard for regression) to avoid a whole load of unnecessary complexity. For classification, what happens if you set "supports_soft_scoring" to false in the scanner? Does it fail?

akumar-cs:
> I think what you've done makes a lot of sense (force soft for classification, hard for regression) to avoid a whole load of unnecessary complexity. For classification, what happens if you set "supports_soft_scoring" to false in the scanner? Does it fail?

@cdraper-CS it won't fail since we return `scores` along with `predictions` and in-case of soft_scoring enabled in scanner for instance shap explanation it would just use the scores. Also verified by running the scan example in the `income_prediction` example

cdraper-CS (approved):

PR #84, opened 2021-01-07
3053 soft h2o part1

cdraper-CS:
First pass of support, which uncovers some issues see https://github.com/CognitiveScale/certifai/issues/3053

Because of those issues, I haven't made it always use soft scoring.  Needs a follow-up PR to at least address the naming issues.


akumar-cs (approved):
this looks good. predict and soft predict are separated and used as model sdk expects setting the labels using h2o methods

PR #86, opened 2021-02-01
Fixed notebook for chnage to scores field names: overall fairness -> …

sdraper-CS:
…fairness

See https://github.com/CognitiveScale/certifai/issues/3182


cdraper-CS (approved):
Notebook runs ok and the bar charts still seem to correspond to the expected results so lgtm

PR #88, opened 2021-02-04
Support requirements.txt for scikit prediction svc

cdraper-CS:
Copied over the relevant parts from v2 (I hope) to support requirements.txt for scikit prediction svc

### DIFF
--- a/models/containerized_model/template.py
+++ b/models/containerized_model/template.py
@@ -178,6 +178,10 @@ def main():
             'prediction_service.py': {
                 'exec_permission': False,
                 'kwargs': {}
+            },
+            'requirements.txt': {
+                'exec_permission': False,
+                'kwargs': {}
             }
         })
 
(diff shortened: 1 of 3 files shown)


models/containerized_model/template.py
```diff
@@ -178,6 +178,10 @@ def generate_python(file_metadata):
             'prediction_service.py': {
                 'exec_permission': False,
                 'kwargs': {}
+            },
+            'requirements.txt': {
```
akumar-cs, line 182: this expects `requirements.txt` file to be present in the templates directory and without which will throw an error. I'll update this pr to add the file

akumar-cs (approved):
added the req file. works now

models/containerized_model/template.py
cdraper-CS: Thanks! I must have missed that in the commit

PR #92, opened 2021-04-22
V2 template

akumar-cs:
its a big-merge so i guess merging and checking might be right way to go about it
though i have done some basic testing


cdraper-CS:
xgboost-iris tests dont work (prediction values need fixing in scan def/metadata) - I'll push a commit when I've finished testing but right now I'm having a series of baffling issues e.g.
```
+ certifai scan -f /Users/cdraper/projects/cortex-certifai-examples/models/containerized_model/examples/h2o_auto_insurance/explain_def.yml
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/bin/certifai", line 5, in <module>
    from certifai.client.cli import main
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/certifai/client/cli.py", line 3, in <module>
    from certifai.client import (scan_cli,
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/certifai/client/scan_cli.py", line 13, in <module>
    from certifai.scanner.certifai_scan import certifai_scan
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/certifai/scanner/certifai_scan.py", line 25, in <module>
    from certifai.common.utils.encoding import CatEncoder
  File "/Users/cdraper/opt/miniconda3/envs/daimodel/lib/python3.6/site-packages/certifai/common/utils/encoding.py", line 5, in <module>
    from sklearn.preprocessing import StandardScaler, OneHotEncoder
ImportError: cannot import name 'StandardScaler'
```

cdraper-CS:
@akumar-cs   I've changed the iris examples to test that we can set the labels for soft predicts. Had to change python prediction service to do so - please check you're ok with the changes.

Also realize that proxy service model URL is set  via an environment variable so this isnt going to be something a user can change in Scan Manager.  I think this means there will have to be a deployment template per prediction service. This may not be a bad starting point as I guess there may also need to be secrets.

cdraper-CS (approved):

akumar-cs:
For proxy, @cdraper-CS scan manager accepts a url via UI and it updates the copy of the default template deploys it so I beleive users can update it. In essence even though it's an env variable it's not exactly used right now in that sense

akumar-cs (commented):
Changes look good for the iris labels

cdraper-CS:
@akumar-cs great - up to you re when to merge then :-)

PR #93, opened 2021-05-12
adding buckets to age

achepur-cs:
Update Patient readmission trust scan notebook - https://github.com/CognitiveScale/certifai/issues/3448


cdraper-CS:
Please you add a sentence explaining that we need to create buckets because age is an integer field with values between 5 and 95. Also ideally rerun this notebook (sorry know that takes a while) and the results notebook so that the notebook state is up-to-date with the changes.

cdraper-CS (approved):

PR #97, opened 2021-05-28
Move notebook TOC down into folder and tweak readmes

cdraper-CS:
Have ToC consistently in the subdirectory. Point only to the about page in docs, not into specific pages. Add a placeholder section in toplevel readme for scan manager folder.


smichalski-cs (approved):
looks good. Thanks for doing this.

PR #99, opened 2021-06-11
Removed Scan section for the german_credit_scanner_definition.yaml

jpathak-CS:
Issue for the PR: https://github.com/CognitiveScale/certifai/issues/3019


jpathak-CS:
> The ticket was intended to change the sample reports that are built into the toolkit (i.e., the ones that are in the certifai repo, not cortex-certifai-examples), although changing it in the examples makes sense too. Have you checked all of the notebooks that use this definition, though? They may have text that tells the user to point the console at local_reports. There may also be text in the public documentation in the cortex-certifai repo that does likewise.

For the cortex-certifai-examples, made changes to definitions(default german-credit), and also altered the comments where reference to local_reports is being made.

notebooks/multiclass/MulticlassHeartDisease.ipynb
```diff
@@ -286,7 +286,7 @@
    "source": [
     "# Run the scan.\n",
     "# By default this will write the results into individual report files (one per model and evaluation\n",
-    "# type) in the 'local_reports' directory relative to this notebook. This may be disabled by specifying\n",
+    "# type) in the 'reports' directory relative to this notebook. This may be disabled by specifying\n",
```
akumar-cs, line 289: Is the location changed? From local reports to reports

akumar-cs (approved):

notebooks/multiclass/MulticlassHeartDisease.ipynb
jpathak-CS: @akumar-cs Without any path mentioned the default location would be /reports local to the notebook.

akumar-cs: Ok

PR #102, opened 2021-07-22
Fix: Prediction Service Deployments fail when the namespace is number

nsanika-CS:
Covers [Prediction Service Deployments fail when the namespace is in all numbers #3616](https://github.com/CognitiveScale/certifai/issues/3616)


pkandarpa-cs:
how does this fix the issue?

nsanika-CS:
> how does this fix the issue?

@pkandarpa-cs wrapping between quotes explicitly sets namespace as a string, in Current version even if you convert numbers to string and pass to render, jinja stills renders it as number, thats why wrapping between quotes to make this as a string type.

cdraper-CS (approved):

mchutani-CS (approved):

PR #103, opened 2021-08-05
Added tutorial for remote scan using Python APIs

jpathak-CS:
Issue: https://github.com/CognitiveScale/certifai/issues/3731

Created a notebook to perform scan using remote scan using Python APIs.
Used default german_credit_definition.yaml.


cdraper-CS:
We really need to test this against Pro Azure. I think that is using Certifai 1.3.6. I strongly suspect some of the options (preflight, output) etc may not exist in that version of Pro and so it may fail. 

We probably also need to test this with both the 1.3.6 client libraries AND the latest client libraries as we don't know which Prudential will be using.

Also, can we change "Namespace" class to something like "Options"? I found it quite confusing as I assumed it was a k8s namespace object, but it isn't.

In general for examples, it's better if we use the CatEncoder from certifai.common. Not a blocker for this though.

jpathak-CS:
> We really need to test this against Pro Azure with corresponding client libraries as thats what we claim as supported (we tell customer to use client that you download from Pro). I think that is using Certifai 1.3.6. It's possible some of the options (e.g., preflight) may not exist in that version of Pro and so it may fail.
> 
> It would be useful but at this point not essential to know if 1.3.11 talking to 1.3.6 works so we have that info in case they need to update their client.
> 
> Also, can we change "Namespace" class to something like "Options"? I found it quite confusing as I assumed it was a k8s namespace object, but it isn't.
> 
> In general for examples, it's better if we use the CatEncoder from certifai.common. Not a blocker for this though.

Tested with 1.3.10 as well as 1.3.4,
Keeping changes needed for 1.3.10 after discussion with @cdraper-CS.

pkandarpa-cs (approved):

cdraper-CS (approved):

tutorials/remote_scan_tutorial/RemoteScanTutorial.ipynb
```diff
@@ -0,0 +1,968 @@
...
+     "output_type": "stream",
+     "text": [
+      "Created certifai scan ID aa15ea4f59c8\n"
+     ]
+    }
+   ],
+   "source": [
+    "from certifai.client.remote.remote_scan import remote_scan\n",
+    "scan_file = \"scan_definition.yaml\"\n",
+    "list_args = Options(alias=remote_alias, definition_file=scan_file, config=[], cpu_req=None, details=False, dry_run=False, info=False, mem_req=None, model=None, model_id_tags=None, output=None, parallel=0, preflight=False, report=None)\n",
```
laguirre-cs, line 892: Noticed the `preflight` argument here. We don't support preflight for remote scans. The operator can run the preflight variant but we don't allow the `--preflight` flag at the CLI level. There shouldn't be anything stopping the user from running a preflight scan here.

tutorials/remote_scan_tutorial/RemoteScanTutorial.ipynb
```diff
@@ -0,0 +1,968 @@
...
+    "remote_scan(parsed_args=list_args, args=[])"
```
laguirre-cs, line 893: The `args` parameter should be the list equivalent to `parsed_args` - this is an example of how these functions are closely tied with the CLI. The argument list is an attribute of the scan job, and the operator expects to parse those strings to create the container args.

So, if you set a value in the `Options` instance and don't set the equivalent in `args` list **it won't be reflected in the remote job**.

laguirre-cs (commented):
We should update the table in the README.md with an entry for this tutorial.

laguirre-cs:
@jpathak-CS

Checked with @cdraper-CS, my review comments were from the perspective of this being used by a wider audience. So you don't need to address my comments. I do suggest keeping my last comment in mind though if we write any supplementary material.

PR #104, opened 2021-10-13
Add an example notebook running counterfactual sampling with and without model access

laguirre-cs:
Issue: https://github.com/CognitiveScale/certifai/issues/3910


notebooks/patient_readmission/README.md
```diff
@@ -19,3 +20,9 @@ Certifai to analyze the trust scores for the models:
 
 4. Run a scan to evaluate the trust scores for the models using the [patient-readmission-trust-scan.ipynb](patient-readmission-trust-scan.ipynb) notebook, and view the results in the Certifai Console.
 5. Investigate the trust scores in the [patient-readmission-trust-results.ipynb](patient-readmission-trust-results.ipynb) notebook.
+
+After at lest completing steps 1 and 2 above, follow the steps below to use
```
cdraper-CS, line 24: least

cdraper-CS:
In the scan notebook...

Need to run the second notebook as well as the first, to get the scan definition. 

Under 'Loading the Explanations Reports' this sentence has gone awry.

```This notebook assumes that it is being run directly after the previous notebook. was did not have access to the model.```

Under 'Extract the Explanations' drop `We'll then print out the first two for the logit and mlp models.` as you have this in the next section and we only have logit model with one prediction.

cdraper-CS:
If we merge this before 1.3.13, we should add info about how to set the early access flag.   We should probably also say 'Needs Certifai 1.3.13, or 1.3.12 with flag set...'

cdraper-CS:
@laguirre-cs  Yes, merge to prerelease after resetting prerelease to master. Then remember we have to merge that!

PR #109, opened 2022-04-06
Example was failing due to added outcome validation

cdraper-CS:
Changed `train.py` so it generates datasets with the transformed 1 or 0 values for readmitted.

Added `app_test.py`.

Removed unused code in `app.py`.


models/patient_readmission/app.py
```diff
@@ -15,75 +15,6 @@
...
-                             encoder=encoder,
-                             host='0.0.0.0',
-                             supports_soft_scores=False
-                             )
-    # Production mode requires Certifai 1.3.6 or higher
-    app.run(production=True, log_level='warning', num_workers=3)
-
 
 if __name__ == '__main__':
     main()
```
laguirre-cs, line 20: There is no longer a `main` function... I'm really not sure if the app setup should be in within the 'main' block or not. I don't think it matters unless we're using gunicorn or something similar...

laguirre-cs (approved):

models/patient_readmission/app.py
cdraper-CS: I removed the main, but we are using gunicorn in production mode. I do have a qualm vs the original main (see import of CleanPipeline) - I'm remembering now about having to jump through hoops to get imports used in the pickle to work. But it runs ok so now I'm wondering whether later gunicorn has fixed it or what?

laguirre-cs: I don't remember/know what those problems were.. There are subtleties like class definitions can't be pickled, so the definition needs to be available when un-pickling.

Reason why gunicorn came to mind is that the workers are not running in the main thread (so gotta watch what code is/isn't called by main) - but I think that doesn't matter here because the gunicorn workers shouldn't import this module

cdraper-CS: The model's running in the workers, so would need CleanPipeline, though.

PR #114, opened 2022-09-22
Move Scan Manager setup documentation to README and re-use wording from docs site

laguirre-cs:
Changes:
* Moves `setup.md` to `README.md`
* Updates wording based on content from our docs site - https://github.com/CognitiveScale/cortex-certifai/pull/269

Separated this from #113 - no reason to hold up the script going in while we work on the corresponding doc.

Related to issue: https://github.com/CognitiveScale/certifai/issues/4603
Related PR: https://github.com/CognitiveScale/cortex-certifai-examples/pull/113


mchutani-CS (approved):

laguirre-cs:
@mchutani-CS  @smichalski-cs FYI- 

I was having trouble with the doc updates because the content is a bit all over the place (in terms of what the containerized model templates are, how to build them, how to setup scan manager with them, etc.).. 

A lot of it feels irrelevant or too detailed/low level, for what I think we would expect an SRE performing the installation to care about when doing the Scan Manager setup with default artifacts. I'm pretty sure they will not be generating new model templates or base image from scratch, but the instructions are all in the framing of "creating new base images".

I added a section for the minimal steps if you're only using the pre defined artifacts -  https://github.com/CognitiveScale/cortex-certifai-examples/pull/114/commits/0c4e6503bd23b808e73def51bcf516435d2c124a

---

There is still the issue of duplication between the contents of this repo & our docs site. I can address part of that by linking off to the github repo in https://github.com/CognitiveScale/cortex-certifai/pull/269.

(If there are any objections content-wise, we can always update the README. We merged this in because it has sufficient instructions for getting an RC built).

PR #118, opened 2022-10-07
added service account for model templates

mchutani-CS:
Relates to: https://github.com/CognitiveScale/certifai/issues/4642


laguirre-cs (approved):

laguirre-cs (commented):
Does there need to be an update in Scan Manager to template the `SERVICE_ACCOUNT_NAME` in the deployment files? (The other PR added it to the config, but I don't know if it'll actually reach the model - I might have the wrong mental model for how the 2 pr's were related)

PR #128, opened 2022-10-26
Model filepath as optional parameter

mchutani-CS:


laguirre-cs (approved):

mchutani-CS:
@laguirre-cs Don't we need pre-release branch to exist? I know we tend to delete branches after merge :P

PR #129, opened 2023-01-06
Remove python 3.6 support, re-run all example notebooks with Python 3.8

laguirre-cs:
Overview of changes:
* Remove references to Python 3.6 (mostly comments).
* Added automation in `build.sh` to run example notebooks & tutorials with configurable python version. The examples will be run on each merge to `pre-release` branch of this repo.
   - The only exclusions are those which involve more setup, namely: AzureML, SageMaker, target_encoded (also uses AzureML), and `tutorials/remote_scan_tutorial`. I will open a separate pull request for these examples, and the multiclass example (because of https://github.com/CognitiveScale/certifai/issues/4694)
* Ran all notebook examples with Python 3.8, and did a quick read through to make sure the notebooks still make sense and had no major/unexpected changes.
* Updated c12e-ci builder image to latest used by Certifai
* You can refer to the commit messages to see how an individual example were changed.

Review wise, I would mostly suggest looking at `build.sh`. A quick look through the notebooks on github would be nice, but I haven't seen any surprising changes in so far.

Issue: https://github.com/CognitiveScale/certifai/issues/4678


notebooks/patient_readmission/patient-readmission-train.ipynb
```diff
@@ -1016,7 +1073,7 @@
    "source": [
     "X = df.drop('readmitted',axis=1)\n",
     "y = df['readmitted']\n",
-    "X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.20, random_state=0)\n",
+    "X_train, X_test, y_train, y_test = train_test_split(X.values, y.values, test_size=0.20, random_state=0)\n",
```
laguirre-cs, line 1076: A side effect of re-running w/ Python 3.8 is that `scikit-learn>=1.0` was used. In version 1.0+ there is a warning emitted by models when you mix pandas/numpy types between training & predictions. For example: 
  ```
   X does not have valid feature names, but StandardScaler was fitted with feature names
  ```
    The fix was to tweak the notebooks to use `numpy` types throughout and/or use Certifai's builtin the `CatEncoder` (in `certifai.common.utils.encoding`), instead of the duplicate copies in this repo (e.g. `notebooks/utils/cat_encoder.py`).  This applied to multiple notebooks, but didn't have a significant change.

pkandarpa-cs (approved):

pkandarpa-cs:
Do you also want to look into setting up a PR pipeline for these tests? Will DM you the details

build.sh
```diff
@@ -121,49 +154,138 @@ function testModels() {
...
+}
+
+function _azuremlModelHeadersDemo() {
+  # TODO: azureml_model_headers_demo
+  true
+}
+
+function _targetEncodedAzuremlNotebook() {
+  # TODO: target_encoded
+  conda remove -n certifai-azure-model-env --all -y
```
pkandarpa-cs, line 247: curious why we are removing this env in this bash func

laguirre-cs: The reason why is that this example has it's conda environment (named `certifai-azure-model-env`), so when re-running the example we should clean/recreate the env.

**This example is not automated in this PR, so this is just some extra/filler code for the moment**.  I've split off the execution of this notebook (and others using azureml/sagemaker) to separate PRs. See: #131 (and #130, #132)

PR #133, opened 2023-02-01
Perform docker build (not push) in PR pipeline and other build script updates

laguirre-cs:
Overview of changes:
* Ignore `artifacts/` directory in `.gitignore` - useful when running build script locally
* Adds help message to the `build.sh` script with all options. I tried to structure the code in a way to avoid spammy output when running `./build.sh
* Adds `local-docker` option to `build.sh` script, this step will also be run in the PR pipeline

(I originally had this as part of #129, but separated it out to avoid more cascading changes with the other feature branches)

### DIFF
--- a/cortex-certifai-examples.gocd.yaml
+++ b/cortex-certifai-examples.gocd.yaml
@@ -47,6 +47,8 @@ common:
               mkdir -p artifacts
               cp -f certifai_toolkit.zip artifacts/
               c12e-ci
+              pip install jinja2 requirements-parser
+              ./build.sh local-docker
   snykScan: &snykScan
     clean_workspace: false
     jobs:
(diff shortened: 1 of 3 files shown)


cortex-certifai-examples.gocd.yaml
```diff
@@ -47,6 +47,8 @@ common:
               mkdir -p artifacts
               cp -f certifai_toolkit.zip artifacts/
               c12e-ci
+              pip install jinja2 requirements-parser
+              ./build.sh local-docker
```
laguirre-cs, line 51: The `c12e-ci` step runs the examples in the pipeline, but it previously wasn't building the base images for prediction service templates. I've added the step to perform build step as a sanity check

pkandarpa-cs (approved):

mcriscolo-cs (approved):

PR #135, opened 2023-03-02
Merge pre-release into master - v1.3.16 changes

laguirre-cs:
Leaving as a draft until we are ready for release.


mchutani-CS:
@laguirre-cs I saw `certifai-examples-rc` pipeline. We should talk about this. I'm not clear on how this fits in the release cycle. Is it to make sure examples run successfully in the `master` branch?

laguirre-cs:
> @laguirre-cs I saw `certifai-examples-rc` pipeline. We should talk about this. I'm not clear on how this fits in the release cycle. Is it to make sure examples run successfully in the `master` branch?

Yes, that's exactly what it is for. The pipeline is triggered by commits to the master branch, and it uses the latest toolkit from the `certifai-toolkit-rc` pipeline.

It was done as part of this issue: https://github.com/CognitiveScale/certifai/issues/4612. We already have a PR pipeline testing on-going changes during dev cycle, however, the toolkit it uses is from `certifai-toolkit` pipeline. That pipeline alone isn't useful for QA to assert the examples pass, because it isn't using the release candidate.

(FYI - GoCD failed to parse the YAML I added in my last change https://github.com/CognitiveScale/cortex-certifai-examples/pull/134, so the `cortex-certifai-examples-rc` pipeline hasn't been run yet. I should have a fix up shortly for this)

mchutani-CS (approved):

PR #136, opened 2023-03-03
Fixes errors from GoCD because of YAML anchors and aliases

laguirre-cs:
RE Issue: https://github.com/CognitiveScale/certifai/issues/4612


laguirre-cs:
Merging this in to get the pipelines working

PR #139, opened 2023-03-09
Add broken link detection as part testing examples and fix broken links

laguirre-cs:
Uses https://github.com/tcort/markdown-link-check to test for broken links.


mchutani-CS (approved):

build.sh
```diff
@@ -180,11 +184,40 @@ function _build_template() {
...
+
+  # Python alternative -> https://github.com/linkchecker/linkchecker - supports markdown but intended for HTML or urls
+  # However this requires a .linkcheckerrc file (with markdown plugin) -> linkchecker -f .linkcheckerrc --check-extern .
+  if ! type "markdown-link-check" > /dev/null;
+  then
+    echo "********************"
+    echo "markdown-link-check is not installed, it will be installed globally with npm"
+    echo "https://github.com/tcort/markdown-link-check"
+    npm install -g markdown-link-check
+    echo "********************"
```
mchutani-CS, line 207: Are these commands intentionally here?

laguirre-cs: The library isn't pre-installed in the GoCD builder image (the examples use the same builder image as `certifai` does - because it needs conda & `node.` Node is only needed for this package).

The installation needs to happen at some point. I put the `if/else` guard around it to avoid re-installing the tool if it's already present though. I did feel a little hesitant putting the install here, because it's not really nice to install something globally if a human is running this. For the sake of the pipeline it seemed alright.

Is there a different location I should put the install step? We could also put a flag to control whether the install happens (otherwise just thrown an error)

PR #142, opened 2023-04-26
Setup CI/CD pipeline to build Python base image for generating Prediction Services 

laguirre-cs:
Notes:
* The same Dockerfile is used to build the Python3.8 and Python3.9 images 
   - Uses `registry.access.redhat.com/ubi9/ubi-minimal:latest` as the base image. The prebuilt [UBI9 Python image](https://catalog.redhat.com/software/containers/ubi9/python-39/61a61032bfd4a5234d59629e?tag=all) only offers Python 3.9 and brings in extra bloat w/ HIGH security vulnerabilities, so I instead used to use the minimal Image with `conda` to install the package
   - The overall build is similar to how the [Certifai images are built](https://github.com/CognitiveScale/certifai/blob/develop/docs/packaging.md#docker-containers) ; it uses a multi-stage build to avoid `miniconda` being in the final image.
* **I've set a up a new stage in the [cortex-certifai-examples-build Pipeline](https://gocd.dci-dev.dev-eks.insights.ai/go/pipeline/activity/cortex-certifai-examples-build)** to build these images
   - An entirely separate pipeline seemed excessive because this pipeline is already triggered with the same materials
   - I decided to keep these image builds separate from the existing `build` stage of the pipeline, to avoid confusion between the resulting images. 
   - **For this release we are continuing to build existing example model images, but after the model packaging is fully implemented the images in the `build` stage shouldn't be needed**
* I decided to the use the existing public dockerhub repo for the Python base images (https://hub.docker.com/r/c12e/cortex-certifai-model-scikit)
   - I'm not a big fan of the 'scikit' in the name, but it will be installed because of the `cortex-certifai-common` package.
   - A new tagging convention is used to differentiate from other images in the repo

-----

For local testing, you will need to [download the toolkit from GoCD](https://gocd.dci-dev.dev-eks.insights.ai/go/pipeline/activity/certifai-toolkit) and place it in the `artifacts/` folder. Then run:

```bash
SKIP_CONDA=true ./build.sh local-docker-builder
```

The images will build the images (not push), the `artifacts/baseBuildReport.json` should contain:
```json
{
  "python38": "c12e/cortex-certifai-model-scikit:base-py38-1.3.17-19-g66a6ae33-d389e12",
  "python39": "c12e/cortex-certifai-model-scikit:base-py39-1.3.17-19-g66a6ae33-d389e12"
}
```

Issue: https://github.com/CognitiveScale/certifai/issues/4763

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -6,7 +6,7 @@ set -eu
 function setGlobals() {
   set -x
   PUSH_IMAGES=false
-  BUILD_ARM=false
+  BUILD_ARM="${BUILD_ARM:-false}"
   SKIP_CONDA="${SKIP_CONDA:-false}"
   SKIP_TOOLKIT="${SKIP_TOOLKIT:-false}"
   RUN_REMOTE_EXAMPLES="${RUN_REMOTE_EXAMPLES:-false}"
--- a/build.sh
+++ b/build.sh
@@ -20,6 +20,7 @@ function setGlobals() {
   NOTEBOOK_DIR="${SCRIPT_PATH}/notebooks"
   TUTORIALS_DIR="${SCRIPT_PATH}/tutorials"
   BUILD_REPORT_JSON="${ARTIFACTS_DIR}/buildReport.json"
+  BASE_IMAGE_BUILD_REPORT_JSON="${ARTIFACTS_DIR}/baseImageBuildReport.json"
   ENV_FILE="${ARTIFACTS_DIR}/.env"
   AWS_ENV_FILE="${ARTIFACTS_DIR}/.aws_env"
   AZURE_ENV_FILE="${ARTIFACTS_DIR}/.azure_env"
@@ -280,7 +281,7 @@ function buildPredictionServiceBuilderImages() {
   fi
 
   # Write build Report
-  echo "{\"python38\": \"${py38_image}\", \"python39\": \"${py39_image}\"}" > "${BUILD_REPORT_JSON}"
+  echo "{\"python38\": \"${py38_image}\", \"python39\": \"${py39_image}\"}" > "${BASE_IMAGE_BUILD_REPORT_JSON}"
 }
 
 function test() {
--- a/build.sh
+++ b/build.sh
@@ -159,7 +159,7 @@ function _installModelRequirements() {
 
 
 function buildModelDeploymentImages() {
-  # Builds Docker Images for the example Containerized Model Types (Scikit, H2O, Proxy, R). These are images are used
+  # Builds Docker images for the example Containerized Model Types (Scikit, H2O, Proxy, R). These are images are used
   # for the default Model Types in Scan Manager in Certifai Enterprise Version 1.3.16 and earlier.
   #
   # We have to enforce a versioning strategy in the example model templates. Part of the trouble here is that template
@@ -169,7 +169,11 @@ function buildModelDeploymentImages() {
   #
   #   `<version-counter>` is a running count we maintain based on the base image, Python version, & other dependencies
   #
-  # Example: c12e/cortex-certifai-model-scikit:v3-1.3.11-120-g5d13c272
+  # Examples:
+  #   c12e/cortex-certifai-model-scikit:v3-1.3.11-120-g5d13c272
+  #   c12e/cortex-certifai-model-h2o-mojo:v4-1.3.17-19-g66a6ae33
+  #   c12e/cortex-certifai-hosted-model:v4-1.3.17-19-g66a6ae33
+  #   c12e/cortex-certifai-model-r:v4-1.3.17-19-g66a6ae33
   #
   # The main reason for this tagging strategy is that earlier images were tagged with 'v1', v2, etc.
   local certifai_version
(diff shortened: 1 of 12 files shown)


build.sh
```diff
@@ -16,9 +16,9 @@ function setGlobals() {
   TOOLKIT_WORK_DIR="${ARTIFACTS_DIR}/toolkit"
   PACKAGES_DIR="${TOOLKIT_WORK_DIR}/packages"
   TEMPLATES_DIR="${SCRIPT_PATH}/models/containerized_model"
+  BASE_IMAGES_DIR="${SCRIPT_PATH}/models/containerized_model/base_images"
   NOTEBOOK_DIR="${SCRIPT_PATH}/notebooks"
   TUTORIALS_DIR="${SCRIPT_PATH}/tutorials"
-  BUILD_REPORT="${ARTIFACTS_DIR}/buildReport.txt"
```
laguirre-cs, line 21: This was a txt version of of `buildReport.json` that wasn't being used - not needed

models/containerized_model/README.md
```diff
@@ -16,7 +16,7 @@
 - H2O MOJO runtime and license file (for running H2O MOJO models).
 - A base docker image which has all the dependencies at the specific versions that were used when the model was trained.
 - Locally installed:
-    - To generate the template: python, PyYaml, Jinja2, and requirements (`pip install -U Jinja2 PyYAML requirements-parser`)
+    - To generate the template: python, PyYaml, Jinja2, and requirements-parser (`pip install -r requirements.txt`)
```
laguirre-cs, line 19: Easier to manage these deps with a `requirements.txt`

models/containerized_model/base_images/Dockerfile.cortex-certifai-python-model-base
```diff
@@ -0,0 +1,95 @@
...
+  name="cortex-certifai-model-python$PY_VERSION-base" \
+  vendor="CognitiveScale" \
+  version=1 \
+  summary="Cortex Certifai Prediction Servie Base Image" \
+  description="Cortex Certifai Enterprise Prediction Servie Base Image" \
+  com.cognitivescale.license_terms="https://www.cognitivescale.com/legal-information/"
+
+ # Create .bashrc file and folders for Python/Pip that belong to the non-root user
+RUN touch //.bashrc && \
+    chown -R 1001 //.bashrc && chmod u+rwx //.bashrc && \
```
laguirre-cs: As mentioned in the PR description, the images are _similar_ to the [Certifai base images](https://github.com/CognitiveScale/certifai/blob/develop/docs/packaging.md#docker-containers). Because the images are running as non-root user, they do not have permissions to write to most of the filesystem. Unlike the Certifai apps, the user does have sufficient permissions to run `pip install`

mchutani-CS: The [Tekton Task](https://github.com/CognitiveScale/certifai/blob/4729-spike-tekton/model-packaging/tekton-tasks/c6i-build-prediction-service-task.yaml#L21) does create a few files and move things around, but this is done on the shared volume (which should be a k8s volume mount). So theoretically we should be good with permissions but this needs testing. Is there a example docker image that I can use to test this?

laguirre-cs: Yeah, it would be good to test it out to end-to-end. There's some instructions in the PR description to build it locally

> For local testing, you will need to [download the toolkit from GoCD](https://gocd.dci-dev.dev-eks.insights.ai/go/pipeline/activity/certifai-toolkit) and place it in the artifacts/ folder. Then run:
```
SKIP_CONDA=true ./build.sh local-docker-builder
```

build.sh
```diff
@@ -16,10 +17,11 @@ function setGlobals() {
...
   BUILD_REPORT_JSON="${ARTIFACTS_DIR}/buildReport.json"
+  BASE_IMAGE_BUILD_REPORT_JSON="${ARTIFACTS_DIR}/baseImageBuildReport.json"
```
laguirre-cs, line 24: The images built in the new stage are kept in a separate build report

models/containerized_model/templates/python/Dockerfile.ubi9
```diff
@@ -0,0 +1,15 @@
...
+# Copyright (c) 2023. Cognitive Scale Inc. All rights reserved.
+# Licensed under CognitiveScale Example Code License https://github.com/CognitiveScale/cortex-certifai-examples/blob/master/LICENSE.md
+#
+FROM {{BASE_DOCKER_IMAGE}}
+
+COPY requirements.txt /tmp/
+RUN pip install -r /tmp/requirements.txt
+
+COPY src /containerized_model/src
+COPY model /containerized_model/model
```
laguirre-cs: FYI @mchutani-CS - The `src/` and `model/` folders are under the `containerize_model/` folder. Let me know, if you would prefer these to be at the root level

models/containerized_model/templates/h2o_mojo/Dockerfile.ubi9
```diff
@@ -0,0 +1,16 @@
+#
```
laguirre-cs, line 1: We haven't setup the CI/CD to build base images for these model types yet - https://github.com/CognitiveScale/certifai/issues/4775

Added this file for this template the h2o mojo model type is as an extension in the Python model. The templating logic (`models/containerized_model/template.py`), expects this file to exist for all Python based model types, so it was easier to add the file than to edit the logic to ignore the error.

models/containerized_model/templates/python/Dockerfile.ubi9
laguirre-cs: @mchutani-CS  I've updated the Dockerfile's to have sufficient permissions for writing to `/src`, `/model` and `/license` directories. The final prediction service image will use these directories when storing the Prediction Service (py file) and the model.pkl at run time.

PR #146, opened 2023-06-15
Automate Model Examples and remove last references to Scikit-Learn v0.23.2

laguirre-cs:
Issue: https://github.com/CognitiveScale/certifai/issues/4812

Overview of changes:
* Fixes broken link to German Credit UCI Dataset - the content was moved. (Fixes pipeline failures)
* Adds Python based unit tests for running the Prediction Service examples.
* Update `run_test.sh` script, for running containerized model examples locally while **skipping h2o models**
* Update example Scan Manager config file to remove `scikit0.23` reference - replace with `python`
* Bump `xgboost` from `1.2.0` to `1.7.2` - this was required to be able to install `xgboost` on the M1 Mac (v1.2.0 was released in 2020!!)


models/base_test.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                p = subprocess.Popen(command, shell=False, stdout=stdout, stderr=stderr, stdin=subprocess.DEVNULL,
+                                     close_fds=True, text=True)
+                yield
+            except Exception:
+                # WARNING: Killing the subprocess may not kill any workers spawned by the process (e.g. gunicorn!)
+                p.kill()
+                p.wait()
+                capture_output(stdout.name, stderr.name)
+                raise
```
laguirre-cs, line 97: The shutdown functionality for the prediction services is kinda wonky right now, especially when testing. It seemed like the best way to handle it was by using Python's subprocess module & just killing the parent process (and for the sake of our examples, the apps run with the dev server so we never risk leaving zombie workers). 

* The builtin `/shutdown` endpoint only works for `gunicorn` servers (when `production=True`) - this is a known bug in the Model SDK package.
* Not all tests will call the `/shutdown` endpoint - like when running python scans. Nor do we expect all the servers to be run with `production=True`
* Python's `kill()` here will kill the process, but if you are using gunicorn, then the spawned workers are left as zombies

scan-manager/docs/setup_artifacts/deployment/config.yml
```diff
@@ -1,11 +1,13 @@
-scikit_0.23:
-  deployment: scikit_0.23_deployment.yml
+python:
+  deployment: python_deployment.yml
```
laguirre-cs, line 2: I'm replacing the `scikit_0.23` reference with `python`, and adding the current set of base images. We will need to update this when the release time comes (or figure out some way to automate this).

Follow up action would be to update the copy on our Certifai Dev to have `python`

laguirre-cs:
Successful PR Pipeline instance: https://gocd.dci-dev.dev-eks.insights.ai/go/tab/build/detail/cortex-certifai-examples-pr/72/examples/1/Notebooks

